PR #8, opened 2022-01-10
adding api key to http requests

shotah:
This is all based on what I have found on stack overflow. I have not actually tried any level of integration testing yet.

Source of solution:
https://stackoverflow.com/questions/26552149/how-to-set-x-api-key-in-the-header-of-http-get-request

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -19,7 +19,7 @@ jobs:
       uses: actions/setup-java@v2.5.0
       with:
         # maximize compatibility with minecraft images
-        distribution: 'termurin'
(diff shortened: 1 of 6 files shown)


itzg (approved):
Excellent. Later I might swap it out with generic header support, but will get this tagged and released ASAP for you to use.

itzg:
Oh yeah, the Java 8 compiler doesn't know about the release designations. I think 9 started that. 

Hmm, could bump the workflow's Java version to 17 since it's now being told to produce Java 8 output.

shotah:
@itzg I believe this is ready to be merged unless you needed anything else done with it.

PR #13, opened 2022-01-30
Use custom user-agent

ThisIsAreku:
Using the default user-agent of Apache HttpClient frequently cause Cloudflare protected domains to reject the request.


itzg (approved):
Thanks! I was surprised things so far had been happy with the default user agent.

itzg:
I'll get this tagged and update the itzg/minecraft-server image

itzg:
The latest itzg/minecraft-server images now include this change.

PR #15, opened 2022-02-03
Adding support for wild cards in Assert File Exists

shotah:
Story:
- Need to Assert file exists using wild cards for MC jar files that continuously change.

Acceptance Criteria: 
- Allow the use of globing to verify if files exist: */*.html etc.

Notes:
- Using [ANT](https://ant.apache.org/manual/api/org/apache/tools/ant/DirectoryScanner.html) to use globbing.
- Added Unit tests for new Globing support

### DIFF
--- a/src/test/java/me/itzg/helpers/assertcmd/FileExistsTest.java
+++ b/src/test/java/me/itzg/helpers/assertcmd/FileExistsTest.java
@@ -49,8 +49,43 @@ class FileExistsTest {
     });
 
     assertThat(errOut).contains(
-        "fileA does not exist",
-        "fileB does not exist"
+        "does not exist"
+    );
+  }
+
+  @Test
+  void passesWhenGlobFindsAllFiles(@TempDir Path tempDir) throws Exception {
+    Files.createFile(tempDir.resolve("file1"));
+    Files.createFile(tempDir.resolve("file2"));
+
+    final String errOut = tapSystemErr(() -> {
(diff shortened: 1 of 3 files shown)


src/test/java/me/itzg/helpers/assertcmd/FileExistsTest.java
```diff
@@ -53,4 +53,40 @@ void failsWhenSomeMissing(@TempDir Path tempDir) throws Exception {
...
+
+  @Test
+  void passesWhenGlobFindsAllFiles(@TempDir Path tempDir) throws Exception {
+    Files.createFile(tempDir.resolve("file1"));
+    Files.createFile(tempDir.resolve("file2"));
+
+    final String errOut = tapSystemErr(() -> {
+      int exitCode = new CommandLine(new FileExists())
+          .execute(
+              String.format("%s/file*", tempDir)
```
itzg, line 65: This might fail on Windows; however, it seems like Windows is now forward-slash tolerant. If it is sensitive the fix would be:
```suggestion
              String.format("%s%sfile*", tempDir, File.separator)
```

itzg (approved):
Nice work

src/test/java/me/itzg/helpers/assertcmd/FileExistsTest.java
shotah: I was running it in windows and it  wasn't failing.

PR #16, opened 2022-02-05
fix(assert): adjusted fileExists to support relative paths

itzg:
When re-running exists tests in itzg/docker-minecraft-server found that relative paths were never getting matched. Stepping through the ant DirectoryScanner code it appears to be quite complicated and archaic.

This change swaps that out with native PathMatcher usage.


itzg:
@shotah since I also bumped into existing tests no longer working since they used relative paths, that gave me the motivation to try harder and implement with the built-in `Files.walk` and `PathMatcher`.

shotah:
No sweat! Sorry I put you through this!

itzg:
> No sweat! Sorry I put you through this!

No need to be sorry. If you couldn't tell already, I enjoy the challenge of scenarios like this 😃

PR #67, opened 2022-12-03
Add interpolation test

doublej472:
When debugging my own issues with environment variable interpolation I noticed there weren't any tests that used environment variables in interpolation.

This PR adds a basic test that requires the interpolator to use substitute some environment variable for a value, along with updating mockito to 4.9.0 and marking the gradlew file as executable


itzg (commented):
Thanks! This looks great.

While you're here and before I forget, can you remove these chmod's during build:

https://github.com/itzg/mc-image-helper/blob/1745b8ed87900827c6d921357e15f0916e140fe2/.github/workflows/test.yml#L26-L27

https://github.com/itzg/mc-image-helper/blob/1745b8ed87900827c6d921357e15f0916e140fe2/.github/workflows/publish-release.yml#L24-L25

doublej472:
@itzg Updated! Removed those `chmod` entries

itzg (approved):
Thanks for the quick update.

PR #88, opened 2023-01-11
Fixed older Forge installers

Megasware128:
Capture the correct exception, extend it with a status code and add support for the odd mc172 versions


itzg (approved):
Thanks!

FYI the build failure can be ignored because I didn't realize the test reports collector isn't setup properly for outside contributors.

Issue #74, opened 2022-12-20
modrinth: support per project version type declaration


itzg:
Also, allow for projects to be split by whitespace only.

Originally suggested at https://github.com/itzg/docker-minecraft-server/issues/1871

michiruf:
Another  suggestion could be, to specify the specific version of a mod, because if things run properly, you may do not want to update those (I myself am pretty much always specifying explicit versions)

itzg:
Yep, I was thinking/assuming the same

JustMangoT:
Except for ``whitespace`` feature, I think this can be closed

PR #126, opened 2023-01-29
Support specific version ID declaration for Modrinth

### LINKED ISSUE #49: feat(modrinth): support project references qualified by specific version ID
Supports https://github.com/itzg/docker-minecraft-server/issues/1650

image

JustMangoT:
closes #49 

Using colon to declare specific version id of a Modrinth's mod. 
Also, I'm not sure if this is the best solution for this; I checked #74 and there is a suggestion to use ``mod#version``.

### DIFF
--- a/src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
+++ b/src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
@@ -205,7 +205,9 @@ public class ModrinthCommand implements Callable<Integer> {
     private Project getProject(String projectIdOrSlug) {
         return fetch(Uris.populateToUri(
             baseUrl + "/project/{id|slug}",
-            projectIdOrSlug
+            projectIdOrSlug.contains(":") ?
+                    projectIdOrSlug.split(":")[0] :
+                    projectIdOrSlug
         ))
             .userAgentCommand("modrinth")
             .toObject(Project.class)
@@ -226,6 +228,15 @@ public class ModrinthCommand implements Callable<Integer> {
         }
     }
 
+    private Version getVersionFromId(String versionId) {
+        return fetch(Uris.populateToUri(
+                baseUrl + "/version/{id}",
+                versionId
+        ))
+                .userAgentCommand("modrinth")
+                .toObject(Version.class)
+                .execute();
+    }
 
     private String arrayOfQuoted(String value) {
         return "[\"" + value + "\"]";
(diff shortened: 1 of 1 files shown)


JustMangoT:
Btw I noticed this may lead to some kind of unexpected errors such as incorrect version ID or incompatible loader, etc... Should those be handled as well?

itzg:
> Btw I noticed this may lead to some kind of unexpected errors such as incorrect version ID or incompatible loader, etc... Should those be handled as well?

It's not the prettiest, but should already throw FailedRequestException with a 404, so I'm thinking don't add any specific logic yet to handle bad IDs. Is that the behavior you've seen?

src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
```diff
@@ -236,8 +247,14 @@ private Stream<? extends Path> processProject(String projectRef) {
```
itzg, line 247: What you have is really good.
To keep the version splitting local to this method, perhaps you could do the split here with limit 2.

src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
```diff
@@ -236,8 +247,14 @@ private Stream<? extends Path> processProject(String projectRef) {
 
         final Project project = getProject(projectRef);
```
itzg, line 248: with that you can unconditionally pass parts[0]

src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
```diff
@@ -236,8 +247,14 @@ private Stream<? extends Path> processProject(String projectRef) {
...
         if (projectsProcessed.add(project.getId())) {
-            final List<Version> versions = getVersionsForProject(project.getId());
-            final Version version = pickVersion(versions);
+            final Version version;
+            if (projectRef.contains(":")) {
```
itzg, line 251: and here conditional on parts.length == 2

itzg (commented):
Thanks so much for working on this one. It'll be very useful for people. 

I'm good to merge, but was wondering if you could consider my suggestion?

JustMangoT:
I'm not sure, I just said that because I was thinking of adding another layer to check if the version belongs to the project since the API routes only requires the version ID.

itzg:
Yeah, let's keep it easy for now and just let it throw as is.

src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
JustMangoT: Should I handle the length > 3 as well or just fallback to the default method?

itzg: The cool thing is if you use [this one](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#split(java.lang.String,int)) you can limit the split to no more than 2 elements.

JustMangoT:
image
tested

itzg (approved):
Thanks!

itzg:
Release is ready to be picked up in docker-minecraft-server https://github.com/itzg/mc-image-helper/releases/tag/1.24.19

PR #127, opened 2023-01-30
Allow per project version type declaration

JustMangoT:
Allow the project version type to be declared using the same method as the version id.
To be honest, this code looks like spaghetti so I will mark this a draft PR.


itzg (commented):
The implementation seems good to me. Have you been able to do some manual testing with it?

JustMangoT:
> The implementation seems good to me. Have you been able to do some manual testing with it?

Yes, already.
image

itzg:
Thanks for the fix. Ready for final review/merge?

JustMangoT:
Yes, I think so. It worked perfectly with my tests, I guess that's enough

itzg (approved):

Issue #271, opened 2023-07-27
Modrinth: Support unpublished Modrinth packs


kMaiSmith:
# Context

I use the Modrinth `.mrpack` format to hand role modpacks for distributing privately, this has proven a very helpful for keeping me and my friends synchronized with the modpack for our server.  However, since i am not able to publish this modpack to modrinth due to licensing issues i would like to be able to point the Minecraft Docker image to my `.mrpack` url and have the modpack be installed to spec.

This backs [itzg/minecraft-docker-server#2144](https://github.com/itzg/docker-minecraft-server/issues/2144)

# Proposed solution

I need this feature for my local service instance, so I intend to attempt to separate the Modrinth pack downloading code from the pack installation code and leverage the existing HTTP classes to facilitate HTTP downloads of `.mrpack`.

I am not sure this calls for changing the interface to the Modrinth part of the CLI, so to leverage this functionality when implemented you would specify the Modrinth Project to be a URL to your `.mrpack`.   The Modrinth installer logic would infer from URLs ending in `.mrpack` to do a direct fetch rather than an API lookup.

## Notes

@itzg Please advise if you think a better approach could be taken

I expect this to take me some time to implement

itzg:
That overall approach sounds good. What were your thoughts about identifying when the installation of the mrpack is up to date? Would you be hosting the file on a web server that supports https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since ?

If you quickly find the changes will be overwhelming, then just let me know. This is a feature I was personally interested in implementing fairly soon anyway.

kMaiSmith:
I had not yet fully considered the how to avoid redundant .mrpack downloads.  I did notice the Modrinth API based fetch flow avoiding redundant downloads based on pack version metadata; given the lack of metadata available before fetching the pack the If-Modified-Since header would work well and be straightforward.

It could also be configured to respect a simple documented pattern, say `<slug>.<version>.mrpack`; if the modpack file matches this pattern the version and modpack name could be extracted and tracked a la Modrinth API modpacks.

One question i have so far to support implementation, does `http.SharedFetch` support fetching from multiple hosts from a single instance of `SharedFetch`?

itzg:
[...]

PR #272, opened 2023-07-29
Add support for downloading and installing informal modrinth packs

kMaiSmith:
Backs #271

## Purpose

Decouple the Modrinth `.mrpack` fetching mechanism from the pack installation mechanism to enable installing packs beyond the Modrinth API 

## Strategy

 - Extract all business logic captured in the `InstallModrinthModpackCommand` class into its own package, leaving the `InstallModrinthModpackCommand` a thin wrapper for the new package interface
 - Separate modpack fetching logic into the `ModrinthAPIPackFetcher` class backed by the `IModrinthPackFetcher` interface
 - Add `ModrinthHttpPackFetcher` to fetch non-Modrinth-API packs also implementing the `IModrinthPackFetcher`
 - Add ModrinthPackInstaller` class to capture `.mrpack` extraction logic 
 - Add new `ModrinthModpack` base class which is responsible for coordinating the fetch and install actions
 - Lots of tests

## Checklist

 - [x] Modrinth API modpack fetcher class fetches `.mrpack` from Modrinth API
 - [x] Modrinth modpack installation/extraction class
 - [x] Refactor `InstallModrinthModpackCommand` to use new logic
 - [x] Add Http modpack fetcher
 - [x] Migrate modpack code back to `modrinth` package


itzg:
I'm sorry, these changes are way too disruptive. You'll need to be extremely minimal in your refactoring. 

Specifically,
- Do not create another package
- Do not name things with "I" prefix. I dislike that convention and it is not a modern naming convention

kMaiSmith:
I felt the need for restructure to get my head around the code for tests, but now with the tests i can comfortably undo many of code-wise refactors.

> Do not create another package

I will re-merge these changes into the original package and see how i can minimize restructuring.  

I will propose a class structure to minimize structural changes:

 - Utilize the `InstallModrinthModpackCommand` to act as the primary logic facilitator (in stead of the new `ModrinthModpack`)
 - Utilize the existing data fields on the `InstallModrinthModpackCommand` to pass around configurations to the fetcher and installer class
 - Move the `ModrinthApiPackFetcher` (+ Interface) and `ModrinthPackInstaller` into the `modrinth` package

> Do not name things with "I" prefix

Noted.  What are the modern practices for this?  Would I just name the interface without the `I`?

---

I can tell I'm not quite getting the Reactive Streams just yet; I'm in the middle of giving the docs a read-through, and i would appreciate feedback on the way the `ModrinthApiPackFetcher` was put together especially in regards to the streams processing and Java best practices.

I expect I will be leaving the `ModrinthPackInstaller` logic mostly untouched as it should apply regardless of the source of the pack, but i do feel compelled to pursue extracting it so i can write some tests around it.

---

I appreciate and welcome any additional feedback, i do not aim to be disruptive. :smile:

itzg:
Sorry for my less-than-constructive initial feedback. It didn't help that I only had a few minutes at the time to reply via the mobile app. I also apologize the total lack of existing unit tests for that -- I was going to suggest I write those before you proceed. Please refer to `src/test/java/me/itzg/helpers/fabric/FabricLauncherInstallerTest.java` to see how I have been doing command/integration level unit tests without imposing extra design layers. Keep in mind, this is only a helper tool and over-designing it is a code smell in this case.

I prefer the sound of your revised strategy. In fact, I am expecting a majority or all of the changes to reside in `InstallModrinthModpackCommand`. Your goal is to introduce as few changes as possible to get from the picocli option for the mrpack file/URL and call the existing `me.itzg.helpers.modrinth.InstallModrinthModpackCommand#processModpackZip`.

As for avoiding "I" prefix. Yes, just name interfaces sensibly and intuitively. It should take a prefix to identify it's an interface...and in most cases the object oriented design should make interface/class non-important to the user/caller.

For the reactive calls, I recommend glancing through https://projectreactor.io/docs/core/release/reference/ and/or https://projectreactor.io/learn.

src/main/java/me/itzg/helpers/modrinth/ModrinthPackFetcher.java
```diff
@@ -0,0 +1,9 @@
+package me.itzg.helpers.modrinth;
+
+import java.nio.file.Path;
+
+import reactor.core.publisher.Mono;
+
+public interface ModrinthPackFetcher {
```
itzg, line 7: Much better name.

src/main/java/me/itzg/helpers/modrinth/ModrinthApiPackFetcher.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private ProjectRef modpackProjectRef;
+    private Mono<Version> modpackVersionStream;
+
+    private Loader modLoaderType;
+    private String gameVersion;
+    private VersionType defaultVersionType;
+    private Path modpackOutputDirectory;
+
+    ModrinthApiPackFetcher(InstallModrinthModpackCommand config) {
```
itzg, line 24: Please avoid passing around the "command" object. Think of it like a REST controller/resource that is the glue between the outside caller and the inner code. Instead, break out the individual parameters and pass those to this constructor.

src/main/java/me/itzg/helpers/modrinth/ModrinthApiPackFetcher.java
```diff
@@ -0,0 +1,82 @@
...
+        this.apiClient = new ModrinthApiClient(
+            config.baseUrl, "install-modrinth-modpack",
+            config.sharedFetchArgs.options());
```
itzg, line 27: Don't create a new instance of this. There should be one for the whole command-call -- the pass the existing one in much like dependency injection like Spring, etc.

src/main/java/me/itzg/helpers/modrinth/ModrinthApiPackFetcher.java
```diff
@@ -0,0 +1,82 @@
...
+        this.gameVersion = config.gameVersion;
+        this.defaultVersionType = config.defaultVersionType;
+        this.modpackOutputDirectory = config.outputDirectory;
+        this.modpackProjectRef =
+            ProjectRef.fromPossibleUrl(config.modpackProject, config.version);
+        this.modpackVersionStream = resolveModpackVersion(this.modpackProjectRef);
```
itzg, line 33: No need to build a field here. Just create and use the stream in `fetchModpack`

itzg (commented):
Just wanted to pass along a few initial code comments.

src/main/java/me/itzg/helpers/modrinth/ModrinthPackInstaller.java
itzg, line 1: In theory I like that you refactored out this class; however, it makes it very difficult for me to see what was new code you introduced vs what was just brought over from `InstallModrinthModpackCommand`. Glancing at it, is the majority the latter case: just refactored the methods over as-is?

src/main/java/me/itzg/helpers/modrinth/ModrinthPackInstaller.java
```diff
@@ -0,0 +1,221 @@
...
+@Slf4j
+public class ModrinthPackInstaller {
+    private final ModrinthApiClient apiClient;
+    private final Path zipFile;
+    private final Path outputDirectory;
+    private final Path resultsFile;
+    private final boolean forceModloaderReinstall;
+    private final SharedFetchArgs sharedFetchArgs;
+
+    public ModrinthPackInstaller(InstallModrinthModpackCommand config, Path zipFile) {
```
itzg, line 40: ...for completeness, again, don't pass in the "command object". Pass the parameters specifically needed and pass in the pre-built api client rather than building a new one.

src/main/java/me/itzg/helpers/modrinth/ModrinthPackInstaller.java
```diff
@@ -0,0 +1,215 @@
...
+@Slf4j
+public class ModrinthPackInstaller {
+    private final ModrinthApiClient apiClient;
+    private final Path zipFile;
+    private final Path outputDirectory;
+    private final Path resultsFile;
+    private final boolean forceModloaderReinstall;
+    private final SharedFetchArgs sharedFetchArgs;
+
+    public ModrinthPackInstaller(ModrinthApiClient apiClient, SharedFetchArgs sharedFetchArgs, Path zipFile, Path outputDirectory, Path resultsFile, boolean forceModloaderReinstall) {
```
itzg, line 40: I was just thinking, rather pass in `SharedFetchArgs` it would be cleaner to pass in `SharedFetch.Options`.

itzg:
The latest changes are looking really good. Thanks for jumping into this feature BTW.

src/main/java/me/itzg/helpers/modrinth/ModrinthPackInstaller.java
kMaiSmith: Everything is as-is with one exception: this code ultimately returns the necessary components to create a `ModrinthMandifest` since you need slug and version to put it together in full. the installer very deliberately does not care about the modpacks Modrinth API metadata.

I will be wrapping this code with a moderate amount of happy path characterization tests next; hopefully those can deliver enough confidence this is still functioning correctly.

src/main/java/me/itzg/helpers/modrinth/ModrinthPackInstaller.java
kMaiSmith: This part confuses me: the only consumer of this in the whole class is the Quilt loader installer; why does the Quilt loader need it while Forge and Fabric don't? (I acknowledge this question is for my curiosity and desire for architectural consistency; having spent so much time in this code now, i want to come back to this class and clean it up but that falls beyond the scope of what i am trying to achieve here)

itzg: Fair question 🙂. I have been evolving the design over time, so Quilt is the newer of the three implementations with cleaner shared fetch usage. If you're interested in a little extra cleanup, the `SharedFetch.Options` should really be passed in rather than the default used here from the `Options.builder().build()` call.

https://github.com/itzg/mc-image-helper/blob/580e62b42f01cc38296e1abd221ec2737ff59da6/src/main/java/me/itzg/helpers/fabric/FabricLauncherInstaller.java#L48

itzg: ...and Forge installer is "pre shared fetch", so it's calling `me.itzg.helpers.http.Fetch#fetch` to create one-off instances.

src/test/java/me/itzg/helpers/modrinth/TestProjectRef.java
```diff
@@ -0,0 +1,90 @@
...
+public class TestProjectRef {
+    private String expectedSlug = "test_project1";
+    private String expectedVersionName = "test_version1";
+    private ProjectRef projectRefUT;
+
+    @Test
+    void testProjectRefHasProjectSlug() {
+        projectRefUT = new ProjectRef(this.expectedSlug, null);
+
+        assertEquals(expectedSlug, projectRefUT.getIdOrSlug());
```
itzg, line 18: Please switch to https://assertj.github.io/doc/ rather than using the junit assertions. It'll be consistent with the rest of the project and they are much more expressive.

kMaiSmith:
@itzg If the unit tests are anything to go by the state of this code should pull down a mrpack from an arbitrary URL.  I still need to:

- migrate the tests to use the assertThat assertions
- piece together how to test this manually and in main docker image

Once these two items are complete i will consider this PR ready for formal review.  Do you have any documentation for the "blessed" method of integrating this into the Docker image itself for testing?

itzg:
I actually had to look that up earlier today, but https://github.com/itzg/docker-minecraft-server/blob/master/docs/misc/contributing/development.md#using-development-copy-of-mc-image-helper is the procedure I use to pre-test the integration.

itzg:
Not as thorough tests as yours, but thought I'd share this feature I just added to the CurseForge commands

https://github.com/itzg/mc-image-helper/pull/277

kMaiSmith:
Just got it tested out and i was able to use it to download my self-hosted modpack.  Due to the nature of my pack i would prefer not to share its URL for validation testing; if necessary I can assemble another one and share its URL with you.

What are the next steps for getting this merged?

itzg:
It seems like unit tests will be sufficient, so don't worry about providing an additional mrpack URL for further manual testing. 

From here the only remaining step is code review and iterating on changes from that.

itzg:
Sorry, I needed to put a fix in for `me.itzg.helpers.modrinth.InstallModrinthModpackCommand#extractOverrides`. I can work on resolving the conflict in your branch if needed.

src/test/java/me/itzg/helpers/modrinth/InstallModrinthModpackCommandTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        int commandStatus = commandUT.call();
+
+        assertEquals(0, commandStatus);
+
+        String actualFileData =
+            new String(Files.readAllBytes(tempDir.resolve(relativeFilePath)));
+
+        assertEquals(expectedFileData, actualFileData);
+    }
```
itzg, line 75: You still need to convert to AssertJ. They have some very nice `Path` (file content, etc) assertions:

https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractPathAssert.html

itzg (changes requested):
Looks great overall! Very nice refactorings. Please switch to AssertJ -- I provided a specific comment where it would have helped even more so.

itzg:
Released in https://github.com/itzg/mc-image-helper/releases/tag/1.33.6

src/main/java/me/itzg/helpers/modrinth/InstallModrinthModpackCommand.java
```diff
@@ -103,271 +68,55 @@ public class InstallModrinthModpackCommand implements Callable<Integer> {
...
-                .flatMap(project ->
-                    apiClient.resolveProjectVersion(
-                            project, projectRef, loader != null ? loader.asLoader() : null, gameVersion, defaultVersionType
-                        )
-                        .switchIfEmpty(Mono.defer(() -> Mono.error(new InvalidParameterException(
-                            "Unable to find version given " + projectRef)))
-                        )
-                        .doOnNext(version -> log.debug("Resolved version={} from projectRef={}", version.getVersionNumber(), projectRef))
-                        .publishOn(Schedulers.boundedElastic()) // since next item does I/O
-                        .filter(version -> needsInstall(prevManifest, project, version))
```
itzg, line 153: @kMaiSmith I just realized you seem to have removed this "force synchronize" feature. Can you create a follow up PR to put it back?

itzg: Created https://github.com/itzg/mc-image-helper/issues/290 to track this

itzg:
@kMaiSmith I need to revert this PR. It has cause a couple of regressions one of which is critical since it prevents users from restart their container with a previously installed modpack.

itzg:
...I will revisit this over the weekend to see if I can fix the versionId being null in the resulting manifest.

drewgingerich:
Hi there! I'm curious about the status of this PR. I'm looking to share an informal modrinth pack by storing the `.mrpack` file in a GitHub repo, and then giving the URL to the minecraft docker container via env var to fetch and install. Going through a whole formal modpack release process on modrinth feels like overkill since I don't want to share it publicly.

As an aside, the existing feature to auto-fetch modpacks from modrinth is awesome. I'm coming back to minecraft after a few years and I'm amazed at how slick self-hosting a modded minecraft server is nowadays!

itzg:
This MR is merged and ready to use in the image. 

The only thing left to implement is handling local mrpack files, but your use case is covered with hosting in GitHub repo.

itzg:
...looks like I need to fix the docs at https://docker-minecraft-server.readthedocs.io/en/latest/types-and-platforms/mod-platforms/modrinth-modpacks/#modpack-project

itzg:
...and I misspoke, the final integration using mrpack URLs may not have been completed.

drewgingerich:
Thanks for taking a look! I was wondering if it had slipped through the cracks since it was merged and then reverted, and GitHub doesn't really make that state visible.

itzg:
FYI changes in this PR were restored in #293

PR #300, opened 2023-08-22
Match numbers in curseforge slugs

pikajude:
Slugs can contain digits, e.g. [this mod](https://www.curseforge.com/minecraft/mc-mods/supermartijn642s-core-lib), although I assume the first character can't be a digit


itzg (approved):
Thanks! I hadn't found any documented requirements on the slugs, so had taken a guess 🙂

itzg:
I'm about to make a change to mc-image-helper shortly, so I'll cut a release after that and bring your change into the server image with that.

itzg:
This is included in this image build https://github.com/itzg/docker-minecraft-server/actions/runs/5946367157

PR #369, opened 2023-12-25
Use proxyWithSystemProperties in SharedFetch

nulano:
Similar to https://github.com/itzg/mc-image-helper/pull/20

This change allows the container to use my proxy server when downloading fabric server if I add (not sure which of these is used by the tool, but IIRC both are needed):
```yaml
environment:
  # use proxy server: 10.0.16.7:3128
  HTTP_PROXY: "http://10.0.16.7:3128"
  JAVA_TOOL_OPTIONS: "-Dhttps.proxyHost=10.0.16.7 -Dhttps.proxyPort=3128 -XX:+PerfDisableSharedMem"
```

### DIFF
--- a/src/main/java/me/itzg/helpers/http/SharedFetch.java
+++ b/src/main/java/me/itzg/helpers/http/SharedFetch.java
@@ -55,6 +55,7 @@ public class SharedFetch implements AutoCloseable {
             .build();
 
         reactiveClient = HttpClient.create(connectionProvider)
+            .proxyWithSystemProperties()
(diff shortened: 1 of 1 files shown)


itzg (approved):
Thanks! I'll get this into the next helper release and image build. Might be a few days but let me know if you need it sooner.

nulano:
> Might be a few days but let me know if you need it sooner.

No rush, I've made a release in my fork and replaced the helper locally, so I won't need it until I update my server again.

PR #371, opened 2024-01-01
Fix exception in FilenameExtractor.filenameFromContentDisposition

alexstaeding:
This PR modifies the method `filenameFromContentDisposition` in `FilenameExtractor` to return null instead of throwing an exception if a filename cannot be extracted from the header.

This allows a filename to still be determined by response path if not correctly determined by the header.

### DIFF
--- a/src/main/java/me/itzg/helpers/http/FilenameExtractor.java
+++ b/src/main/java/me/itzg/helpers/http/FilenameExtractor.java
@@ -45,7 +45,7 @@ public class FilenameExtractor {
                 try {
                     return Rfc5987Util.decode(m.group(2), m.group(1));
                 } catch (UnsupportedEncodingException e) {
-                    log.debug("Failed to decode filename* from {}", headerValue);
+                    log.warn("Failed to decode filename* from {}", headerValue);
                     return null;
                 }
             }
--- a/src/main/java/me/itzg/helpers/http/FilenameExtractor.java
+++ b/src/main/java/me/itzg/helpers/http/FilenameExtractor.java
@@ -6,7 +6,6 @@ import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import lombok.extern.slf4j.Slf4j;
-import me.itzg.helpers.errors.GenericException;
 import org.apache.cxf.attachment.Rfc5987Util;
 import org.apache.hc.client5.http.HttpResponseException;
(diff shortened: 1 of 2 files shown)


src/main/java/me/itzg/helpers/http/FilenameExtractor.java
itzg, line 1: It looks like whitsepace changes made the diff much larger than it needs to be. Can you revert and touch only the minimal lines?

itzg (changes requested):

src/main/java/me/itzg/helpers/http/FilenameExtractor.java
```diff
@@ -24,69 +24,71 @@ public class FilenameExtractor {
...
+        final String filenameStar = parsed.getParameter("filename*");
+        final String filename = parsed.getParameter("filename");
+        if (filenameStar != null) {
+            final Matcher m = RFC_5987_ENCODED.matcher(filenameStar);
+            if (m.matches()) {
+                try {
+                    return Rfc5987Util.decode(m.group(2), m.group(1));
+                } catch (UnsupportedEncodingException e) {
+                    log.debug("Failed to decode filename* from {}", headerValue);
+                    return null;
```
itzg, line 49: I think this scenario should be fatal. The filename value was non-null and it's trying to do the further processing on it.

itzg (changes requested):
As discussed in Discord, the white space changes are fine.

src/main/java/me/itzg/helpers/http/FilenameExtractor.java
itzg: Just changed my mind, but make it a warning log. Since it's beyond the user's control what the server provides, it does seem fair to be non-fatal and best effort continue from there.

alexstaeding: Fixed

itzg (approved):
The unit tests are probably sufficient for this to be safely merged, if you're good with that.

itzg:
Speaking of unit tests, sorry I didn't look earlier, but I need you to add a test case to

https://github.com/itzg/mc-image-helper/blob/master/src/test/java/me/itzg/helpers/http/OutputToDirectoryFetchBuilderTest.java

alexstaeding:
Sorry for the delay. The new test case fails without the other changes in this PR - so it should be good to go.

itzg:
Thanks for wrapping that up!

PR #430, opened 2024-05-25
cf: allow for trailing slash on CF modpack page URL

itzg:
For https://github.com/itzg/docker-minecraft-server/issues/2847


badwolfunit:
Thanks

PR #434, opened 2024-06-07
Support optional --use-temp-dir in get to download to temp dir and copy, to avoid partial jar if interrupted

potomato:
Hi,

This is the first draft of small change to download server jar to temp folder to address my [issue](https://github.com/itzg/docker-minecraft-server/issues/2841).

Some brief notes
- I've tried to keep it small
- Currently it only deals with the single file download. I've not added anything to the OutputToDirectoryHandler yet. Not sure if this is worthwhile but very happy to for consistency
- I've got negative tests for the new switch plus a test that it works when specifying a temporary directory
- The successful test doesn't actually verify it uses the download directory. This is doable with a file system watcher so I'd be happy to do it if you think it's worthwhile.

I'm not a java dev so very happy to receive any requests for changes from tiny to massive.

Thanks.

### DIFF
--- a/src/main/java/me/itzg/helpers/get/GetCommand.java
+++ b/src/main/java/me/itzg/helpers/get/GetCommand.java
@@ -134,6 +134,11 @@ public class GetCommand implements Callable<Integer> {
     @Option(names = "--retry-delay", description = "in seconds", defaultValue = "2")
     int retryDelay;
 
+    @Option(names = {"--use-temp-dir"},
+        description = "Specifies the name of a directory to use for downloading to a temporary file, before it is copied to the final destination",
+        paramLabel = "DIR")
+    Path tempDir;
+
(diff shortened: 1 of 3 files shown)


itzg:
Cool. I'll look more at the code later when I'm at a computer. 

My initial thought though is that I'd prefer it use the option to download to a temp file in the same directory where the final file will be located. I'm thinking it's like how chrome will name the downloading file the real name with ".download" (or something like that) appended to it. The challenge with a temp directory is identifying where a good one would be within the container -- I think I have encountered cases where /tmp surprisingly wasn't viable.

potomato:
Happy to switch it around as needed. Let me know the best approach in due course and I'll revise it.

itzg:
Please go ahead switch to an adjacent temp file rather than the temp directory approach.

PR #486, opened 2024-10-24
feat : add custom highlight to add color to logs

IronMine:


itzg (commented):
Cool, thanks. 

Can you see if logback has helper methods to query if current context is a terminal that supports colors?

itzg (commented):
Switch to using the `withJansi` feature here

https://logback.qos.ch/manual/layouts.html#coloring

IronMine:
This is excatly what i am trying to do but when it's set to true it does not have color anymore ^^

IronMine:
this does not color the logs ! when i desactivate withJansi they are colored back

itzg:
It looks like you need to use their highlight and not your custom one.

IronMine:
i tried... no with theyr highlight it does not work too

itzg:
It worked fine for me; however, I did have to run it in a regular terminal rather than within IntelliJ's run config (in my case). This what I ran to test the changes:

```
./gradlew installDist
build/install/mc-image-helper/bin/mc-image-helper install-curseforge --output-directory=/tmp/cf --slug=prominence-2-classic
```

and this is what I have for the pattern:
```
            <pattern>%highlight([mc-image-helper] %d{HH:mm:ss.SSS} %-5level : %msg%n)</pattern>
```

itzg:
...and just now added a "test-logging-level" subcommand to further confirm things:

image

itzg:
> okay but outside of intellIj terminal right ?

I'm running the above commands from IntelliJ Terminal window, but not via the Gradle/app run config

image

IronMine:
> > okay but outside of intellIj terminal right ?
> 
> I'm running the above commands from IntelliJ Terminal window, but not via the Gradle/app run config
> 
> image
> 
Okay nice it works for me ! Can you confirm it works with the custom hilight ( so the warnings are yellow) ?

itzg:
> Can you confirm it works with the custom hilight ( so the warnings are yellow) ?

I don't want to use custom highlighting if their built-in works -- besides yours is still missing the is-terminal conditional. I would prefer to have it not match colors than reinvent their wheel.

IronMine:
as you want. But i do not reinvent their wheel, i just customize the colors of the highlight --

itzg:
I can't see that they're conditionalizing either, so I'm fine with your custom one then

Double checked there's at
https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/pattern/color/HighlightingCompositeConverter.java

itzg (approved):
Thanks!

IronMine:
> Kicked off a release build to get included in image
> 
> https://github.com/itzg/mc-image-helper/releases/tag/1.40.3

Thanks a lot !

PR #567, opened 2025-06-03
Add support for Leaf

gothickitty93:
Added loader type 'leaf'


itzg:
This released in https://github.com/itzg/mc-image-helper/releases/tag/1.42.1 and can be bumped here 

https://github.com/itzg/docker-minecraft-server/blob/1289fea1c66a4491925a8d67ad33ee789514e764/Dockerfile#L51

PR #570, opened 2025-06-08
Include Folia in results file

kingcavespider1:
Install paper hardcodes type to PAPER, not sure if just removing that is the best idea but if you a recommended way of doing it I can implement that


itzg:
I can't remember off hand why that was needed other than ensuring a specific TYPE was available for the image scripts downstream from Modrinth modpack install. Setting it to the requested type, paper, folia, etc seems like it would have been fine. In fact, setting FAMILY seems like it would have been just as helpful.

kingcavespider1:
Didn't want to change ResultsFileWriter too much so I this should just have it set the type to the uppercase of project

PR #571, opened 2025-06-08
Prefix system

### LINKED ISSUE #564: Support plugin qualifier for Modrinth downloads
Like https://github.com/itzg/docker-minecraft-server/issues/3452 but support a plugin prefix. Came up as part of https://github.com/itzg/docker-minecraft-server/pull/3484#issuecomment-2933052683

kingcavespider1:
Been having trouble getting docker to include the file to locally test this, it builds and passes all tests but almost certainly not done in the best way. I was able to test it outside of the docker image with

./mc-image-helper modrinth --loader forge --projects "paper:simple-voice-chat, fabric:lithium, ferrite-core:beta, datapack:terralith, quilt:comforts:6.3.4+1.20.1" --game-version 1.20.1

It downloaded all successfully

Don't think this is ready to merge without heavy cleanup and testing but could use a second opinion on it

Closes #564

### DIFF
--- a/src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
+++ b/src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
@@ -44,9 +44,9 @@ import picocli.CommandLine.Option;
 public class ModrinthCommand implements Callable<Integer> {
 
     public static final String DATAPACKS_SUBDIR = "datapacks";
-    @Option(names = "--projects", description = "Project ID or Slug",
+    @Option(names = "--projects", description = "Project ID or Slug. Prefix with loader: e.g. fabric:project-id",
         split = SPLIT_COMMA_NL, splitSynopsisLabel = SPLIT_SYNOPSIS_COMMA_NL,
-        paramLabel = "id|slug"
+        paramLabel = "[loader:]id|slug"
     )
     List<String> projects;
 
@@ -157,7 +157,7 @@ public class ModrinthCommand implements Callable<Integer> {
         return Manifests.load(outputDirectory, ModrinthManifest.ID, ModrinthManifest.class);
     }
 
-    private Stream<Version> expandDependencies(ModrinthApiClient modrinthApiClient, Project project, Version version) {
+    private Stream<Version> expandDependencies(ModrinthApiClient modrinthApiClient, Loader loader, String gameVersion, Project project, Version version) {
         log.debug("Expanding dependencies of version={}", version);
         return version.getDependencies().stream()
             .filter(this::filterDependency)
@@ -170,7 +170,7 @@ public class ModrinthCommand implements Callable<Integer> {
                     if (dep.getVersionId() == null) {
                         log.debug("Fetching versions of dep={} and picking", dep);
                         depVersion = pickVersion(
-                            getVersionsForProject(modrinthApiClient, dep.getProjectId())
+                            getVersionsForProject(modrinthApiClient, dep.getProjectId(), loader, gameVersion)
                         );
                     }
                     else {
@@ -193,7 +193,7 @@ public class ModrinthCommand implements Callable<Integer> {
                     log.debug("Resolved version={} for dep={}", depVersion.getVersionNumber(), dep);
                     return Stream.concat(
                             Stream.of(depVersion),
-                            expandDependencies(modrinthApiClient, project, depVersion)
+                            expandDependencies(modrinthApiClient, loader, gameVersion, project, depVersion)
                         )
                         .peek(expandedVer -> log.debug("Expanded dependency={} into version={}", dep, expandedVer));
                 }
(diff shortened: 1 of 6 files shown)


src/main/java/me/itzg/helpers/modrinth/ProjectRef.java
```diff
@@ -22,10 +25,13 @@ public class ProjectRef {
...
     );
-    private static final Pattern PROJECT_REF = Pattern.compile("(?<datapack>datapack:)?(?<idSlug>[^:]+?)(:(?<version>[^:]+))?");
+    private static final Set<String> VALID_LOADERS = Arrays.stream(Loader.values())
+        .map(Enum::name)
+        .map(String::toLowerCase)
+        .collect(Collectors.toSet());
 
     private final String idOrSlug;
-    private final boolean datapack;
+    private final Loader loaderRef;
```
itzg, line 34: Sorry for the nitpick, but the "Ref" in "loaderRef" keeps throwing me off. Can you name this and the variables just "loader"?

itzg:
> Been having trouble getting docker to include the file to locally test this

You might have found this already https://docker-minecraft-server.readthedocs.io/en/latest/misc/contributing/development/#using-development-copy-of-tools but I have to admit it's not easiest testing workflow to get working initially. 

Your testing approach with direct command invocation is a lot of what I do anyway during development of mc-image-helper changes.

src/main/java/me/itzg/helpers/modrinth/ProjectRef.java
```diff
@@ -36,32 +42,48 @@ public class ProjectRef {
     private final String versionNumber;
 
     public static ProjectRef parse(String projectRef) {
-        final Matcher m = PROJECT_REF.matcher(projectRef);
-        if (!m.matches()) {
-            throw new InvalidParameterException("Invalid project reference: " + projectRef);
+        // First, try to split into potential loader prefix and the rest
+        final int firstColon = projectRef.indexOf(':');
```
itzg, line 46: Since the delimiters are always colons, then could use `String.split` and test for the length of the resulting array to determine the different scenarios. Your current approach is totally fine though.

src/main/java/me/itzg/helpers/modrinth/ProjectRef.java
```diff
@@ -36,32 +42,48 @@ public class ProjectRef {
...
+        Loader loaderRef = null;
+        String rest = projectRef;
+        
+        if (firstColon > 0) {
+            final String prefix = projectRef.substring(0, firstColon);
+            if (VALID_LOADERS.contains(prefix.toLowerCase())) {
+                loaderRef = Loader.valueOf(prefix);
```
itzg, line 53: Probably should be this since the enum names are all lowercase too

```suggestion
                loaderRef = Loader.valueOf(prefix.toLowerCase());
```

src/test/java/me/itzg/helpers/modrinth/ModrinthCommandTest.java
```diff
@@ -311,7 +311,7 @@ void handlesDatapacksSpecificVersion(boolean absoluteWorldDir, @TempDir Path tem
                     tempDir.resolve(worldDir).toString()
                     : worldDir,
                 "--game-version", "1.21.1",
-                "--loader", "datapack",
```
itzg, line 314: Thanks, that must have been a typo originally, but masked by the "datapack:" prefix resolution.

itzg (commented):
Thanks. A couple minor tweaks and it seems good to me.

src/main/java/me/itzg/helpers/modrinth/ProjectRef.java
kingcavespider1: Yeah I have no idea what that was about, it was loader first but that broke something so I changed it. Seems to work now, I changed how I was going about this about 3 times since starting

src/main/java/me/itzg/helpers/modrinth/ProjectRef.java
kingcavespider1: Yeah that whole section was the last I got working, I had it working with a slightly modified version of the original regex but it couldn't cope with giving it slug:version without a loader prefix in front

kingcavespider1:
Alright that seems like it works for datapacks now

src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
itzg, line 1: For some reason it thinks the whole file is one big diff.

kingcavespider1: Weird let me see if I can figure that out

kingcavespider1: Yeah that's what I think too I'll try to see if I can figure that out, it seemed to work normally before

kingcavespider1: What os and editor do you use? or line ending settings?

itzg: I develop both on Mac and Windows. Currently on Mac it's opening that file with "LF" line ends and I confirmed my git config is set the same as

https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings

It'll be more about how that particular commit got converted or not when pushed. Let me try re-pushing that file since locally it shows fine.

kingcavespider1:
Whitespace changes were most likely just one of the tools I used deciding it should reformat everything on saving. Usually I'd notice when the diff looks bigger than I expect but gets easier to miss with bigger changes.

src/main/java/me/itzg/helpers/modrinth/ModrinthCommand.java
kingcavespider1: Haven't really been thinking about what mine was set to until now because it usually seemed to figure itself out, but think LF would be the default for me. easy to forget about EOL and text encoding because it usually just works until it ends up breaking something

itzg:
Released in https://github.com/itzg/mc-image-helper/releases/tag/1.43.0

PR #580, opened 2025-06-19
Add Modrinth project files support

fireph:
Uses the same syntax as Curseforge: `@/path/to/modrinth-mods.txt`. Added tests to make sure it is working as expected. I tried to follow how it was done for Curseforge as closely as possible to keep things consistent. I will make a separate PR to update the docs at itzg/docker-minecraft-server/docs/mods-and-plugins/modrinth.md

docs PR: https://github.com/itzg/docker-minecraft-server/pull/3506

### DIFF
--- a/src/test/java/me/itzg/helpers/modrinth/ModrinthCommandTest.java
+++ b/src/test/java/me/itzg/helpers/modrinth/ModrinthCommandTest.java
@@ -397,13 +397,12 @@ class ModrinthCommandTest {
         final int exitCode = new CommandLine(
             new ModrinthCommand()
         )
-            .setExpandAtFiles(false)
             .execute(
                 "--api-base-url", wm.getRuntimeInfo().getHttpBaseUrl(),
                 "--output-directory", tempDir.toString(),
                 "--game-version", "1.19.2",
                 "--loader", "fabric",
-                "--projects", "@" + listingFile
+                "--projects=@" + listingFile
             );
 
         assertThat(exitCode).isEqualTo(ExitCode.OK);
--- a/src/test/java/me/itzg/helpers/modrinth/ModrinthCommandTest.java
+++ b/src/test/java/me/itzg/helpers/modrinth/ModrinthCommandTest.java
@@ -386,9 +386,9 @@ class ModrinthCommandTest {
 
         final Path listingFile = Files.write(tempDir.resolve("listing.txt"),
             Arrays.asList(
-                "fabric-api:vNBWcMLP",
+                "fabric-api",
                 "# This is a comment",
-                "cloth-config:qA00xo1O",
+                "cloth-config",
                 "",
                 "# Another comment"
             )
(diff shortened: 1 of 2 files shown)


src/test/java/me/itzg/helpers/modrinth/ModrinthCommandTest.java
```diff
@@ -378,6 +380,37 @@ void handlesDatapacksLatestVersion(@TempDir Path tempDir) {
...
+        )
+            .execute(
+                "--api-base-url", wm.getRuntimeInfo().getHttpBaseUrl(),
+                "--output-directory", tempDir.toString(),
+                "--game-version", "1.21.5",
+                "--loader", "fabric",
+                "--projects", "@" + listingFile
+            );
+
+        assertThat(exitCode).isEqualTo(ExitCode.OK);
```
itzg, line 408: The builds in GitHub don't seem to reveal the actual failure, so I'll see if it fails for me locally too

https://github.com/itzg/mc-image-helper/actions/runs/15746318998/job/44383891924 Add Modrinth project files support · itzg/mc-image-helper@d844d66

itzg (commented):
Thanks! 

Just need to investigate the unit test failure.

src/test/java/me/itzg/helpers/modrinth/ModrinthCommandTest.java
```diff
@@ -378,6 +380,38 @@ void handlesDatapacksLatestVersion(@TempDir Path tempDir) {
...
+                "cloth-config",
+                "",
+                "# Another comment"
+            )
+        );
+
+        final int exitCode = new CommandLine(
+            new ModrinthCommand()
+        )
+            .setExpandAtFiles(false)
```
fireph, line 400: If anyone stumbles upon this in the future, this is needed when the argument is a named argument otherwise it will always expand the file.

itzg: This makes sense; however, I'm now wondering why I didn't need it for `me.itzg.helpers.curseforge.CurseForgeFilesCommandTest#usingListingFile`

fireph: apparently it is because it does not expand @filename when it is the primary arg, but it does when it is a named arg

itzg: Must be. Their docs don't mention that behavior, so that's odd

https://picocli.info/#AtFiles

I wonder if that will also need to be disabled here

https://github.com/itzg/mc-image-helper/blob/master/src/main/java/me/itzg/helpers/McImageHelper.java#L174

Had you tried running the built version?

fireph: I have not tried running this change inside of a docker-minecraft-server, what is the best process to test changes in mc-image-helper when running the docker container?

fireph: ah I see what you mean now. Running the main entry point works as expected:

mc-image-helper-test

-------------------------

Using modrinth-mods.txt:
``` modrinth-mods.txt
fabric-api
# This is a comment
cloth-config

# Another comment
```

src/test/java/me/itzg/helpers/modrinth/ModrinthCommandTest.java
```diff
@@ -378,6 +380,38 @@ void handlesDatapacksLatestVersion(@TempDir Path tempDir) {
...
+            .execute(
+                "--api-base-url", wm.getRuntimeInfo().getHttpBaseUrl(),
+                "--output-directory", tempDir.toString(),
+                "--game-version", "1.19.2",
+                "--loader", "fabric",
+                "--projects", "@" + listingFile
```
itzg, line 406: The need for `@@` is what worried me. That would have been what users of the image would need to also do. Instead, using `--arg=value` syntax seems to bypass the at-file processing of picocli. So, changing to this seems best:

```suggestion
            .execute(
                "--api-base-url", wm.getRuntimeInfo().getHttpBaseUrl(),
                "--output-directory", tempDir.toString(),
                "--game-version", "1.19.2",
                "--loader", "fabric",
                "--projects=@" + listingFile
```

itzg (approved):
Thanks!

itzg:
Now released in https://github.com/itzg/mc-image-helper/releases/tag/1.45.0

Can bump the version at

https://github.com/itzg/docker-minecraft-server/blob/1cf2783daf8c79532df21f3b3e007042061aedeb/Dockerfile#L51

As part of https://github.com/itzg/docker-minecraft-server/pull/3506

PR #581, opened 2025-06-19
feat(users): offline UUD generation support

FedotCompot:
Offline UUID implementation, used if `manage-users` receives `--offline` flag

### DIFF
--- a/src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
+++ b/src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
@@ -14,6 +14,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.Callable;
 import lombok.extern.slf4j.Slf4j;
 import me.itzg.helpers.errors.GenericException;
@@ -25,7 +26,10 @@ import me.itzg.helpers.http.Uris;
 import me.itzg.helpers.json.ObjectMappers;
 import me.itzg.helpers.users.model.JavaOp;
 import me.itzg.helpers.users.model.JavaUser;
+
+import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.maven.artifact.versioning.ComparableVersion;
+
 import picocli.CommandLine.ArgGroup;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.ExitCode;
@@ -45,6 +49,9 @@ public class ManageUsersCommand implements Callable<Integer> {
     @SuppressWarnings("unused")
     @Option(names = {"--help", "-h"}, usageHelp = true)
     boolean help;
+
+    @Option(names = {"--offline"}, required = false, description = "Use for offline server, UUIDs are generated")
+    boolean offline;
 
     @Option(names = "--output-directory", defaultValue = ".")
     Path outputDirectory;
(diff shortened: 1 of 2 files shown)


src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
```diff
@@ -25,7 +29,13 @@
 import me.itzg.helpers.json.ObjectMappers;
 import me.itzg.helpers.users.model.JavaOp;
 import me.itzg.helpers.users.model.JavaUser;
+
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.codec.digest.Crypt;
+import org.apache.commons.codec.digest.Md5Crypt;
```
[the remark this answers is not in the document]
FedotCompot: Sure, didn't have LSP, fixed, removed also unused `Os` and `BigInteger`

src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
```diff
@@ -329,4 +341,38 @@ private void processInputAsFile(SharedFetch sharedFetch, String filePathUrl) thr
     private boolean usesTextUserList() {
         return version != null && new ComparableVersion(version).compareTo(MIN_VERSION_USES_JSON) < 0;
     }
+
+    private static JavaUser getOfflineUUID(String username) {
+        byte[] bytes = new byte[16];
+        try {
+            bytes = MessageDigest.getInstance("MD5").digest(("OfflinePlayer:"+username).getBytes());
+        }catch(NoSuchAlgorithmException e){
+            System.exit(1);
```
itzg, line 350: Would be better to either

1. `log.warn` if continuing without UUID is a decent fallback, per the next line you have
2. throw a `me.itzg.helpers.errors.GenericException` wrapping `e` with explanation

src/test/java/me/itzg/helpers/users/ManageUsersCommandTest.java
```diff
@@ -5,6 +5,7 @@
 import static uk.org.webcompere.modelassert.json.JsonAssertions.assertJson;
 import static uk.org.webcompere.modelassert.json.condition.ConditionList.conditions;
 
+import com.ctc.wstx.io.SystemId;
```
itzg, line 8: Looks like this import is also unused

src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
FedotCompot: The thing is that function is never going to throw, the world where "MD5" algorithm does not exist is not possible, it would be better if there was an alternative that does not throw at all

itzg: Understood but a System.exit just looks gross there. Throw the wrapped exception so it looks right and I have it bail immediately with that already.

src/test/java/me/itzg/helpers/users/ManageUsersCommandTest.java
FedotCompot: Fixed 👌

src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
FedotCompot: Yeah, totally agree, did some quick research, found a solution that does not throw! Way better now

itzg (approved):
Nice!

PR #627, opened 2025-09-04
Alternative to #434: download via .download temp file then rename

pgcudahy:
I have a very slow internet connection, so interrupted downloads have often caused me problems. I saw that #434 from @potomato addressed this but seems to have stalled out. I'm a decent programmer but have no experience with java, so full disclosure that I used claude code to help me with this. I'm confident in the logic in GetCommand.java, but I'm not familiar with the testing framework. The tests look sensible to me and pass.


potomato:
Well done @pgcudahy. I'm no java dev and ran out of time and skills! Your solution is simpler though which is better. 
Glad to see it fixed!

Issue #686, opened 2026-01-02
No obvious way to enable offline mode for manage-users command


Azertinv:
I might be stupid but I don't see a way to enable the `--offline` argument to the `manage-users` command from the compose file, there doesn't seem to be documentation on how to use this flag in #583 

I also saw that the `--offline` flag was redundant with the `<username>:offline` flag in the whitelist/ops format.

I'm currently rolling a patched mc-image-helper in the meantime that just remove the check for this flag and only takes into account the flag from the whitelist.

```diff
diff --git a/src/main/java/me/itzg/helpers/users/ManageUsersCommand.java b/src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
index 6ad4cf2..a57c19a 100644
--- a/src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
+++ b/src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
@@ -245,7 +245,7 @@ public class ManageUsersCommand implements Callable<Integer> {
                 }

                 // User is not online, generating offline UUID
-                if (offline && user.getFlags().contains("offline")) {
+                if (user.getFlags().contains("offline")) {
                     log.debug("Resolved '{}' as offline user", user.getName());
                     // update UUID keeping the other fields in case of existing user
                     return finalUser.setUuid(getOfflineUUID(user.getName()));
```

Here is my current compose.yaml
```yaml
# Generated by setupmc.com on 2026-01-02T21:03:59.828Z
services:
  mc:
    image: itzg/minecraft-server:latest
    container_name: minecraft-server
    pull_policy: daily
    tty: true
    stdin_open: true
    ports:
      - "25565:25565"
    environment:
      EULA: "TRUE"
      TYPE: "FABRIC"
      ICON: "/server-icon.png"
      OVERRIDE_ICON: "true"
      VERSION: "1.21.11"
      MEMORY: "16384M"
      ENFORCE_SECURE_PROFILE: "false"
      ENABLE_ROLLING_LOGS: "true"
      MAX_PLAYERS: "30"
      USE_AIKAR_FLAGS: "true"
      USE_MEOWICE_FLAGS: "true"
      TZ: "Europe/Paris"
      DIFFICULTY: "2"
      SPAWN_PROTECTION: "1"
      MODRINTH_PROJECTS: |-
        fabric-api
        easywhitelist
        easyauth
      OPS: |-
        User1
      ENABLE_WHITELIST: "true"
[...]

PR #687, opened 2026-01-07
Removed the --offline flag in the manage-users command as it was redundant

Azertinv:
Fix for #686


itzg:
Released in https://github.com/itzg/mc-image-helper/releases/tag/1.52.2 

Now just needs to be bumped here

https://github.com/itzg/docker-minecraft-server/blob/9c01462c0ca1e2f2b2bfd5bf83aa2078cb796d6a/Dockerfile#L52

PR #732, opened 2026-03-22
Offline username being hashed twice

KhooHaoYit:
There's a mistake that got introduced in https://github.com/itzg/mc-image-helper/pull/638

The username would get hashed first before being passed to `UUID.nameUUIDFromBytes` where it would get hashed again by the function, causing it to generate the wrong UUID

Ref: [UUID.java#L168-L181](https://github.com/openjdk/jdk/blob/jdk-25%2B36/src/java.base/share/classes/java/util/UUID.java#L168-L181)

### DIFF
--- a/src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
+++ b/src/main/java/me/itzg/helpers/users/ManageUsersCommand.java
@@ -370,13 +370,7 @@ public class ManageUsersCommand implements Callable<Integer> {
     }
 
     private static String getOfflineUUID(String username) {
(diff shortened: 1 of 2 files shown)


itzg (approved):
Thanks

itzg:
Sure would have been nice if the javadoc mentioned that 

https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/UUID.html#nameUUIDFromBytes(byte%5B%5D)

...but that's what I get for not researching what a type 3 UUID is.

PR #748, opened 2026-04-16
Feat: Clean server Jar libraries on server install

### LINKED ISSUE #3821: Is it possible to limit count of downloaded server jars?
### Describe the enhancement

I realized that alongside accumulated logs, old server version jars are also kept indefinitely - but for logs, I see I can now use `ROLLING_LOG_MAX_FILES: "n"`.

Is there a similar env var for server jars?

harrisonablack:
When server jars are ran, they install a set of dependencies to "./libraries". When updating server versions, they may install or update libraries, however they do not remove old libraries.

Implemented a `--clean-libraries` flag which defaults to false. This option checks the required libs from the current server jar against the installed libs in `./libraries` and deletes all installed and not required libraries.

Currently this only supports Paper jars, however support will be added for further TYPES.

Partial implementation to fix itzg/docker-minecraft-server#3821

### DIFF
--- a/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
+++ b/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
@@ -82,7 +82,7 @@ public class LibraryCleaner {
             try {
                 Files.delete(libraryFolder.resolve(s));
             } catch (Exception e) {
-                log.warn("Failed to delete library {} {}", s, e);
+                log.warn("Failed to delete library {}", s, e);
             }
 
         }
@@ -189,12 +189,12 @@ public class LibraryCleaner {
                                 Files.delete(dir);
                             }
                         } catch (IOException e) {
-                            log.error("Failed to delete directory {} {}", dir.toString(), e);
+                            log.error("Failed to delete directory {}", dir, e);
                         }
                     });
 
         } catch (Exception e) {
-            log.error("Failed to walk directory {} {}", root.toString(), e);
+            log.error("Failed to walk directory {}", root, e);
         }
     }
 }
--- a/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
+++ b/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
@@ -57,13 +57,16 @@ public class LibraryCleaner {
         }
 
         final List<String> installedLibraries;
+
+
+        if (!Files.isDirectory(libraryFolder)) {
+            log.debug("Failed to read library folder {}", libraryFolder);
+        }
+
         try {
             installedLibraries = readInstalledLibraries(libraryFolder);
-        } catch (Exception e) {
-
-            // Will fail on fresh install as no libs installed
-            // Therefore no /library created
-            log.debug("Failed to read installed libraries {}", e);
+        } catch (IOException e) {
+            log.warn("Failed to read installed libraries", e);
             return;
         }
 
--- a/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
+++ b/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
@@ -99,16 +99,14 @@ public class LibraryCleaner {
      * @throws IOException
      */
     private List<String> readJarLibraries(Path jarFile, LibraryListPaths libraryListPath) throws IOException {
-        List<String> libs = new ArrayList<String>();
+        final List<String> libs;
 
         try (JarFile serverJar = new JarFile(jarFile.toString());) {
 
             JarEntry libraryList = serverJar.getJarEntry(libraryListPath.getPath());
 
             if (libraryList == null) {
-                log.error("Failed to read library list {} for server type {}", libraryListPath.getPath(),
-                        libraryListPath.name());
-                return libs;
+                throw new IOException("Failed to read library list for server jar");
             }
 
             try (
@@ -137,9 +135,7 @@ public class LibraryCleaner {
 
             }
 
-        } catch (Exception e) {
-            log.error("Failed to read server jar {} {}", jarFile.toString(), e);
-        }
+        } 
 
         return libs;
     }
--- a/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
+++ b/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
@@ -165,9 +165,10 @@ public class LibraryCleaner {
      */
     private HashSet<String> compareInstalledRequiredLibraries(List<String> installed, List<String> required) {
         HashSet<String> installedNotRequired = new HashSet<String>(installed);
+        HashSet<String> requiredSet = new HashSet<String>(required);
 
         for (String s : installed) {
-            if (required.contains(s)) {
+            if (requiredSet.contains(s)) {
                 installedNotRequired.remove(s);
             }
         }
--- a/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
+++ b/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
@@ -7,6 +7,7 @@ import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
@@ -64,7 +65,7 @@ public class LibraryCleaner {
             return;
         }
 
-        oldLibraries = compareInstalledRequiredLibraries(installedLibraries, requiredLibraries);
+        oldLibraries = new ArrayList<String>(compareInstalledRequiredLibraries(installedLibraries, requiredLibraries));
 
         if (oldLibraries.isEmpty()) {
             return;
@@ -162,8 +163,8 @@ public class LibraryCleaner {
      * @param required  List of libraries required by Server Jar
      * @return List of currently installed libraries not required by the Server Jar
      */
-    private List<String> compareInstalledRequiredLibraries(List<String> installed, List<String> required) {
-        List<String> installedNotRequired = new ArrayList<String>(installed);
+    private HashSet<String> compareInstalledRequiredLibraries(List<String> installed, List<String> required) {
+        HashSet<String> installedNotRequired = new HashSet<String>(installed);
 
         for (String s : installed) {
             if (required.contains(s)) {
--- a/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
+++ b/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
@@ -1,0 +1,178 @@
+package me.itzg.helpers.libraries;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * LibraryCleaner
+ */
+@Slf4j
+@AllArgsConstructor
+public class LibraryCleaner {
+    private static String INSTALLED_LIBRARY_FOLDER = "libraries";
+
+    // @TODO I believe all server jar types save libs to "/libraries" however will
+    // have to verify when implementing for further types
+
+    private final LibraryListPaths libraryListPath;
+    private final Path serverJar;
+    private final Path workingFolder;
+    private final Path libraryFolder;
+
+    public LibraryCleaner(Path serverJar, LibraryListPaths libraryListPath) {
+        this.serverJar = serverJar;
+        this.libraryListPath = libraryListPath;
+
+        this.workingFolder = serverJar.getParent();
+        this.libraryFolder = workingFolder.resolve(INSTALLED_LIBRARY_FOLDER);
+    }
+
+    /**
+     * cleanLibraries reads currently installed libraries against required libraries
+     * for Server Jar, removes non-required libraries
+     */
+    public void cleanLibraries() {
+        List<String> requiredLibraries, installedLibraries, oldLibraries;
+
+        try {
+            requiredLibraries = readJarLibraries(serverJar, libraryListPath);
+        } catch (Exception e) {
+            log.debug("Failed to read server jar libraries", e);
+            return;
+        }
+
+        try {
+            installedLibraries = readInstalledLibraries(libraryFolder);
+        } catch (Exception e) {
+
+            // Will fail on fresh install as no libs installed
+            // Therefore no /library created
+            log.debug("Failed to read installed libraries {}", e);
+            return;
+        }
+
+        oldLibraries = compareInstalledRequiredLibraries(installedLibraries, requiredLibraries);
+
+        if (oldLibraries.isEmpty()) {
+            return;
+        }
+
+        log.info("Removing " + String.valueOf(oldLibraries.size()) + " deprecated installed libraries");
+
+        for (String s : oldLibraries) {
+            try {
+                Files.delete(libraryFolder.resolve(s));
+                deleteEmptyParentDirectories(libraryFolder, libraryFolder.resolve(s));
+            } catch (Exception e) {
+                log.debug("Failed to delete library {} {}", s, e);
+            }
+
+        }
+    }
+
+    /**
+     * Reads required libraries from inside Jarfile Manifest
+     * 
+     * @param jarFile         Path to server Jar
+     * @param libraryListPath Path to library list inside jar
+     * @return List of all libraries required by server jar
+     * @throws IOException
+     */
+    private List<String> readJarLibraries(Path jarFile, LibraryListPaths libraryListPath) throws IOException {
+        List<String> libs = new ArrayList<String>();
+
+        JarFile serverJar = new JarFile(jarFile.toString());
+        JarEntry libraryList = serverJar.getJarEntry(libraryListPath.getPATH());
+
+        InputStream is = serverJar.getInputStream(libraryList);
+        InputStreamReader isr = new InputStreamReader(is);
+        BufferedReader br = new BufferedReader(isr);
+
+        libs = br.lines()
+                .map(String::trim)
+                .map(s -> s.split("\\s+"))
+                .filter(parts -> parts.length > 1)
+                .map(parts -> parts[2])
+                .collect(Collectors.toList());
+
+        serverJar.close();
+        return libs;
+    }
+
+    /**
+     * @param libraryFolder Folder where libraries are currently installed
+     * @return returns list of all individual library Jars currently installed
+     * @throws IOException
+     */
+    private List<String> readInstalledLibraries(Path libraryFolder) throws IOException {
+        List<String> libs = new ArrayList<String>();
+
+        libs = Files.walk(libraryFolder)
+                .filter(Files::isRegularFile)
+                .map(libraryFolder::relativize)
+                .map(Path::toString)
+                .collect(Collectors.toList());
+
+        return libs;
+    }
+
+    /**
+     * Takes the differance of currently installed and required libraries
+     * 
+     * @param installed List of currently installed libraries
+     * @param required  List of libraries required by Server Jar
+     * @return List of currently installed libraries not required by the Server Jar
+     */
+    private List<String> compareInstalledRequiredLibraries(List<String> installed, List<String> required) {
+        List<String> installedNotRequired = new ArrayList<String>(installed);
+
+        for (String s : installed) {
+            if (required.contains(s)) {
+                installedNotRequired.remove(s);
+            }
+        }
+
+        return installedNotRequired;
+    }
+
+    /**
+     * Takes currently delete file, moves upwards in filesystem and deletes any
+     * empty folders
+     * 
+     * @param baseRoot    Root folder where libraries are installed
+     * @param deletedFile Path to recently deleted file
+     */
+    private void deleteEmptyParentDirectories(Path baseRoot, Path deletedFile) {
+        Path current = deletedFile.getParent();
+        while (current != null && !current.equals(baseRoot)) {
+            try (Stream<Path> entries = Files.list(current)) {
+                if (entries.findAny().isPresent()) {
+                    return;
+                }
+            } catch (IOException e) {
+                log.debug("Failed to inspect directory {}", current, e);
+                return;
+            }
+            try {
+                Files.delete(current);
+            } catch (IOException e) {
+                log.debug("Failed to delete empty directory {}", current, e);
+                return;
+            }
+            current = current.getParent();
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


src/main/java/me/itzg/helpers/libraries/LibraryListPaths.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@AllArgsConstructor
+@Getter
+public enum LibraryListPaths {
+    PAPER("META-INF/libraries.list"),
+    FORGE("bootstrap-shim.list");
+
+    private String PATH;
```
itzg: Should be lowercase since it's a field and not enum value.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,178 @@
...
+            return;
+        }
+
+        oldLibraries = compareInstalledRequiredLibraries(installedLibraries, requiredLibraries);
+
+        if (oldLibraries.isEmpty()) {
+            return;
+        }
+
+        log.info("Removing " + String.valueOf(oldLibraries.size()) + " deprecated installed libraries");
```
itzg: This app uses slf4j so use placeholders `{}` instead of string-building

https://www.slf4j.org/faq.html#logging_performance

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,178 @@
...
+
+        for (String s : oldLibraries) {
+            try {
+                Files.delete(libraryFolder.resolve(s));
+                deleteEmptyParentDirectories(libraryFolder, libraryFolder.resolve(s));
+            } catch (Exception e) {
+                log.debug("Failed to delete library {} {}", s, e);
```
itzg: Warn is probably better. Reports an issue but non-fatal

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,178 @@
...
+     * @param jarFile         Path to server Jar
+     * @param libraryListPath Path to library list inside jar
+     * @return List of all libraries required by server jar
+     * @throws IOException
+     */
+    private List<String> readJarLibraries(Path jarFile, LibraryListPaths libraryListPath) throws IOException {
+        List<String> libs = new ArrayList<String>();
+
+        JarFile serverJar = new JarFile(jarFile.toString());
+        JarEntry libraryList = serverJar.getJarEntry(libraryListPath.getPATH());
```
itzg: Probably needs a null check in case it's an invalid server jar.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,178 @@
...
+
+        InputStream is = serverJar.getInputStream(libraryList);
+        InputStreamReader isr = new InputStreamReader(is);
+        BufferedReader br = new BufferedReader(isr);
```
itzg: Need to wrap in try-with-resources to ensure closure

https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,178 @@
...
+     * 
+     * @param jarFile         Path to server Jar
+     * @param libraryListPath Path to library list inside jar
+     * @return List of all libraries required by server jar
+     * @throws IOException
+     */
+    private List<String> readJarLibraries(Path jarFile, LibraryListPaths libraryListPath) throws IOException {
+        List<String> libs = new ArrayList<String>();
+
+        JarFile serverJar = new JarFile(jarFile.toString());
```
itzg: Wrap in try-with-resources

https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,178 @@
...
+        JarEntry libraryList = serverJar.getJarEntry(libraryListPath.getPATH());
+
+        InputStream is = serverJar.getInputStream(libraryList);
+        InputStreamReader isr = new InputStreamReader(is);
+        BufferedReader br = new BufferedReader(isr);
+
+        libs = br.lines()
```
itzg: An inline comment here explaining the expected per-line syntax would be good. Especially so since I'd be surprised if all server types end up following the same syntax -- they usually can't agree on anything 😉

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,178 @@
...
+     * @throws IOException
+     */
+    private List<String> readInstalledLibraries(Path libraryFolder) throws IOException {
+        List<String> libs = new ArrayList<String>();
+
+        libs = Files.walk(libraryFolder)
+                .filter(Files::isRegularFile)
+                .map(libraryFolder::relativize)
+                .map(Path::toString)
+                .collect(Collectors.toList());
```
itzg, line 156: Perhaps also filter on filename suffix of ".jar". Just to be totally sure and it'll be self documenting what this file walk is looking for.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Takes the differance of currently installed and required libraries
+     * 
+     * @param installed List of currently installed libraries
+     * @param required  List of libraries required by Server Jar
+     * @return List of currently installed libraries not required by the Server Jar
+     */
+    private List<String> compareInstalledRequiredLibraries(List<String> installed, List<String> required) {
+        List<String> installedNotRequired = new ArrayList<String>(installed);
```
itzg: Using a HashSet here is slightly faster and self documents the intent.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,178 @@
...
+                return;
+            }
+            try {
+                Files.delete(current);
+            } catch (IOException e) {
+                log.debug("Failed to delete empty directory {}", current, e);
+                return;
+            }
+            current = current.getParent();
+        }
```
itzg: Walk file tree with post traversal makes this easier, such as

```
Files.walkFileTree(startPath, new SimpleFileVisitor<Path>() {
            @Override
            public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOException {
                // This is called after all files and sub-dirs have been visited
                try (DirectoryStream<Path> ds = Files.newDirectoryStream(dir)) {
                    if (!ds.iterator().hasNext()) {
                        Files.delete(dir);
                        System.out.println("Deleted empty directory: " + dir);
                    }
                }
                return FileVisitResult.CONTINUE;
            }
        });
```

or to follow streams usage above

```
import java.io.IOException;
import java.nio.file.*;
import java.util.Comparator;

public void deleteEmptyWithStreams(Path root) throws IOException {
    try (var stream = Files.walk(root)) {
        stream.sorted(Comparator.reverseOrder()) // Deepest directories first
              .filter(Files::isDirectory)
              .forEach(path -> {
                  try {
                      // delete() throws DirectoryNotEmptyException if not empty
                      Files.deleteIfExists(path);
                  } catch (IOException ignored) {
                      // Directory was not empty, skipping
                  }
              });
    }
}
```

itzg (changes requested):
Sorry for the delay. Looks really good. Just a few nit picks.

src/main/java/me/itzg/helpers/libraries/LibraryListPaths.java
harrisonablack: Renamed in a8afc73

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Reworded 95afda2

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Add try catch on initialising JarFile, as constructor throws `IOException` and `SecurityException`. Check JarEntry if null as `.getJarEntry()` returns null if file not found https://docs.oracle.com/javase/8/docs/api/java/util/jar/JarFile.html. Added in 63a6432

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Updated in 2fea3a1

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Wrapped readers 28a5701

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Effectively wrapped the previous resource handling in commit 28a5701, can't combine all instantiation into single try-with-resources as the `JarEntry` doesn't own its resource, and as such doesn't implement `java.lang.AutoCloseable`, added in 871e582

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: When i checked the initial few jars when testing, they did appear to use the same format, however if they don't user the same file location, why would they use the same format?

Add explanation in 142c746

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Added filter that files are suffixed with ".jar" in commit 3b5a96d

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Updated so that lookups / deletions are on the HashSet, left iteration over the list 7b0188b

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: I preferred the stream optiona personally, also added a check for empty directories instead of relying on the try catch 7589720

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
itzg: I should have elaborated. Since this method wasn't named for a specific server type or file listing type being used, the comment about syntax helps for determining future refactoring and specialization.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,205 @@
...
+            return;
+        }
+
+        try {
+            installedLibraries = readInstalledLibraries(libraryFolder);
+        } catch (Exception e) {
+
+            // Will fail on fresh install as no libs installed
+            // Therefore no /library created
+            log.debug("Failed to read installed libraries {}", e);
```
itzg: Ah, don't use placeholders for exceptions though. They are always the last argument and get special permission processing such as stack dump. 

Also, don't log exceptions at debug. They should be warn or error depending on non-fatal or fatal. Or wrap in an informative exception and rethrow. 

This one seems fatal. Definitely rethrow with one of the common exception used already (like IllegalStateException) and let the CLI processing terminate it out.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,205 @@
...
+        this.workingFolder = serverJar.getParent();
+        this.libraryFolder = workingFolder.resolve(INSTALLED_LIBRARY_FOLDER);
+    }
+
+    /**
+     * cleanLibraries reads currently installed libraries against required libraries
+     * for Server Jar, removes non-required libraries
+     */
+    public void cleanLibraries() {
+        List<String> requiredLibraries, installedLibraries, oldLibraries;
```
itzg: Don't declare multiple variables together (even though Java lets you)

Instead, each should be declared final right before their try blocks below.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,205 @@
...
+            installedLibraries = readInstalledLibraries(libraryFolder);
+        } catch (Exception e) {
+
+            // Will fail on fresh install as no libs installed
+            // Therefore no /library created
+            log.debug("Failed to read installed libraries {}", e);
+            return;
+        }
+
+        oldLibraries = new ArrayList<String>(compareInstalledRequiredLibraries(installedLibraries, requiredLibraries));
```
itzg: Doesn't need to be rewrapped in a new array list.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,205 @@
...
+    /**
+     * Reads required libraries from inside Jarfile Manifest
+     * 
+     * @param jarFile         Path to server Jar
+     * @param libraryListPath Path to library list inside jar
+     * @return List of all libraries required by server jar
+     * @throws IOException
+     */
+    private List<String> readJarLibraries(Path jarFile, LibraryListPaths libraryListPath) throws IOException {
+        List<String> libs = new ArrayList<String>();
```
itzg: Declare final

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,205 @@
...
+     * @param libraryFolder Folder where libraries are currently installed
+     * @return returns list of all individual library Jars currently installed
+     * @throws IOException
+     */
+    private List<String> readInstalledLibraries(Path libraryFolder) throws IOException {
+        List<String> libs = new ArrayList<String>();
+
+        libs = Files.walk(libraryFolder)
+                .filter(Files::isRegularFile)
+                .filter(path -> path.getFileName().toString().endsWith(".jar"))
```
itzg: `Path` already has that

```suggestion
                .filter(path -> path.getFileName().endsWith(".jar"))
```

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,205 @@
...
+     */
+    private HashSet<String> compareInstalledRequiredLibraries(List<String> installed, List<String> required) {
+        HashSet<String> installedNotRequired = new HashSet<String>(installed);
+        HashSet<String> requiredSet = new HashSet<String>(required);
+
+        for (String s : installed) {
+            if (requiredSet.contains(s)) {
+                installedNotRequired.remove(s);
+            }
+        }
```
itzg: Sets make this easy

```suggestion
        installedNotRequired.removeAll(required);
```

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,205 @@
...
+     */
+    private List<String> readJarLibraries(Path jarFile, LibraryListPaths libraryListPath) throws IOException {
+        List<String> libs = new ArrayList<String>();
+
+        try (JarFile serverJar = new JarFile(jarFile.toString());) {
+
+            JarEntry libraryList = serverJar.getJarEntry(libraryListPath.getPath());
+
+            if (libraryList == null) {
+                log.error("Failed to read library list {} for server type {}", libraryListPath.getPath(),
```
itzg: Use warn for non-fatal

No need for getPath

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,205 @@
...
+                        .map(String::trim)
+                        .map(s -> s.split("\\s+"))
+                        .filter(parts -> parts.length > 1)
+                        .map(parts -> parts[2])
+                        .collect(Collectors.toList());
+
+            }
+
+        } catch (Exception e) {
+            log.error("Failed to read server jar {} {}", jarFile.toString(), e);
```
itzg: Noting another that needs placeholder removed for exception

No need to for toString. Slf4j handles that.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
```diff
@@ -0,0 +1,205 @@
...
+                            if (files.findAny().isEmpty()) {
+                                Files.delete(dir);
+                            }
+                        } catch (IOException e) {
+                            log.error("Failed to delete directory {} {}", dir.toString(), e);
+                        }
+                    });
+
+        } catch (Exception e) {
+            log.error("Failed to walk directory {} {}", root.toString(), e);
```
itzg: And unneeded exception placeholder 

No need to for toString. Slf4j handles that.

itzg (changes requested):
Just a few last things. Mostly missed from before. 

I highly recommend looking at an slf4j tutorial to help understand my comments there.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: > This one seems fatal. Definitely rethrow with one of the common exception used already (like IllegalStateException) and let the CLI processing terminate it out.

The idea is that on an initial run, the `./libraries` folder created by the server jar won't exist, and as such throws a `FileNotFoundException`. My thought would be to warn and return as it's not fatal in this situation.

Having a general `catch (Exception e)` definitely a bad choice, but instead, could we just have a check if the `libraries` folder exists, and warn if it doesn't?

```suggestion
if (!Files.isDirectory(libraryFolder)) {
    log.warn("Libraries folder doesn't exist");
    return;
}
```

`.isDirectory` returns "false if the file does not exist, is not a directory, or it cannot be determined if the file is a directory or not."

itzg: Disregard the log level comment; however, relying on an exception as an expected case is misleading. Yes, using isDirectory is a proper defensive condition. Debug level log is good then.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Added in 53c0963

Also when i changed to HashSets prior, i changed the return type to `HashSet`, requiring to redeclare `new ArrayList` later, reverted back to returning `List`

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Added in b31f2fb, debug on folder not existing, warn on failure to read

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Changed in 6a5c381

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: I initially wrapped this in a `new ArrayList` as before when i updated the

```java
private List<String> compareInstalledRequiredLibraries(List<String> installed, List<String> required) {
```

function, i changed the return type to `HashSet`, but reverted that change

New arraylist removed in 6a5c381 when changing to final.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
itzg: Or returning `Collection` makes it compatible input for lists or sets.

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: In changing to `final`, throws `Local variable not initialised` on early returns on failure. As the function already throws IOException, i removed early returns and logs, instead throwing errors which will be caught by the try-catch block the function is encased in.

Changed in 344fa3e

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Yeahhh, thats a bit silly

ebe275b

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: As the logs have been refactored to throw exceptions in that function in 344fa3e, i've instead changed the try-catch that function is encapsulated in to log level warn

2b6188d

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Cleaned in a2d0751

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: Cleaned in a2d0751

itzg (approved):
Thanks for addressing all my nitpicks. And thanks for this initial feature.

itzg:
Included in https://github.com/itzg/mc-image-helper/releases/tag/1.57.0

rwjack:
Cool, thanks guys!

So how do I actually make this work on my end?

itzg:
@rwjack wait for @harrisonablack to [apply the upgrade here](https://github.com/itzg/docker-minecraft-server/blob/d34111c3e35ee5f5151ad19aff05c58bb8ee8a45/Dockerfile#L58) and incorporate the new invocation into https://github.com/itzg/docker-minecraft-server

src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
harrisonablack: @itzg I just realised that this commit broke the `LibraryCleaner`.

Because we are performing a `Path.endsWith(String other)` on a `Path` it doesn't perform the same as `String.endsWith(String other)`

"boolean endsWith([String](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html) other) Tests if this path ends with a Path, constructed by converting the given path string, in exactly the manner specified by the [endsWith(Path)](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#endsWith-java.nio.file.Path-) method. On UNIX for example, the path "foo/bar" ends with "foo/bar" and "bar". It does not end with "r" or "/bar"."

That was my bad comiting and not testing the functionality.

Will add a pr to revert the change

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#endsWith-java.lang.String-

harrisonablack: Pr #752

itzg: That's weird. What's the point of them having that method when getFilename() does the same thing? Oh well, thanks for catching that.

itzg: nm, I need to get some sleep 😄

itzg:
@harrisonablack can you make the changes for the image startup scripts before adding more types to cleanup? Users like @rwjack can't even make use of your feature yet.

harrisonablack:
> @harrisonablack can you make the changes for the image startup scripts before adding more types to cleanup? Users like @rwjack can't even make use of your feature yet.

Yep, can add Paper and Purpur support in the startup-scripts before proceeding

PR #753, opened 2026-05-04
modrinth: handle case when a project depends on a user-declared project with non-default version type

itzg:
[Reported in Discord](https://discord.com/channels/660567679458869252/660569641550217327/1500542269596635401)


itzg:
Need to follow up with a unit test for this case. The tricky part is inducing the ordering of project depending on alpha project. Normally a hashmap doesn't guarantee key order.

itzg:
Included in https://github.com/itzg/docker-minecraft-server/pull/4044

PR #757, opened 2026-05-07
Feat: Clean server Jar libraries on Quilt install

harrisonablack:
From what i understand of the quilt installer, it downloads a slim entry jar, which is formatted as `quilt-server-{mc-version}-{quilt-loader-version}.jar`. The launcher then starts `server.jar` to start the server.

From my testing, the naming of the main jar; `server.jar` is consistent across quilt versions, both inside mc-image-helper and outside when installing manually

The `server.jar` contains the LibrariesList, which is why its directly passed to the LibraryCleaner, instead of the `...launcher.jar`

### DIFF
--- a/src/main/java/me/itzg/helpers/quilt/QuiltInstaller.java
+++ b/src/main/java/me/itzg/helpers/quilt/QuiltInstaller.java
@@ -18,6 +18,8 @@ import me.itzg.helpers.files.Manifests;
 import me.itzg.helpers.files.ResultsFileWriter;
 import me.itzg.helpers.http.Fetch;
 import me.itzg.helpers.http.SharedFetch;
+import me.itzg.helpers.libraries.LibraryCleaner;
+import me.itzg.helpers.libraries.LibraryListPaths;
 import me.itzg.helpers.mvn.MavenRepoApi;
 import me.itzg.helpers.versions.MinecraftVersionsApi;
 import org.jetbrains.annotations.Blocking;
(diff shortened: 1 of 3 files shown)


src/main/java/me/itzg/helpers/quilt/QuiltInstaller.java
```diff
@@ -194,6 +199,16 @@ private QuiltManifest runInstaller(String minecraftVersion, Path installerPath,
             throw new GenericException("Failed to rename server launcher", e);
         }
 
+
+        if (cleanLibraries) {
+            final Path serverJar = outputDir.resolve("server.jar");
```
itzg, line 204: These server types that use a launcher-installer probably aren't matching your expectations. Please design out what happens on first install, upgrade, and when user switches to another server type.

itzg (commented):
I'm not sure if these other types are worth addressing. Please provide your plans and justification for each ahead of time.

harrisonablack:
> Please provide your plans and justification for each ahead of time.

Apologies, i just updated the description of the PR as you commented this

itzg:
I wanted to see

1. Helper runs
  1. Drops installer
  2. Cleans up files from server.jar — doesn’t exist
2. Installer runs
  1. Creates and runs server.jar
  2. libraries get populated 
3. User restarts to pickup upgrade
4. Helper runs
  1. Drops new installer
  2. Cleans up files from server.jar — nothing to do since it’s the same one created at 2.1
  3. Cleans up previous installer 
5. New installer runs
  1. Overwrites server.jar
  2. Populates new libraries needed
6. Any restart
7. Helper finally gets a chance to cleanup libraries from 2.2

Is the feature doing what’s expected? Eventually.

itzg:
Ugh it messed up my numbering. I ran out of time to look at this.

PR #758, opened 2026-05-07
Clean server Jar libraries on Purpur install

harrisonablack:
Add picocli flag "--clean-libraries" and cleanLibraries logic similar to that implemented in `PaperInstallCommand` in #748


itzg (approved):
Thanks. Unlike Quilt, this one seems good to address.

PR #759, opened 2026-05-08
ServerLibraryCleaner: Add early return on failure to read Libraries folder

harrisonablack:
Mentioned in itzg/docker-minecraft-server#4046

When a server is *initially* created, and the `--clean-libraries` flag is passed, the `./libraries` folder in the output dir doesn't exist yet.

There was already a check on this, `Files.isDirectory` returns "false if the file does not exist, is not a directory, or it cannot be determined if the file is a directory or not".

```java
if (!Files.isDirectory(libraryFolder)) {
  log.debug("Failed to read library folder {}", libraryFolder);
}

```

However there was no early return statement if it returned false, and instead was propagated to the user as a Warn and exception logged

```java
try {
  installedLibraries = readInstalledLibraries(libraryFolder);
} catch (IOException e) {
  log.warn("Failed to read installed libraries", e);
  return;
}
```

Added a return statement on `Files.isDirectory` returning false

### DIFF
--- a/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
+++ b/src/main/java/me/itzg/helpers/libraries/LibraryCleaner.java
@@ -61,6 +61,7 @@ public class LibraryCleaner {
 
         if (!Files.isDirectory(libraryFolder)) {
             log.debug("Failed to read library folder {}", libraryFolder);
+            return;
         }
 
         try {


harrisonablack:
I considered other approaches, like i mentioned the idea of reading manifests.

However the source of truth if we *__can__* run the LibraryCleaner is if there are libraries currently installed, if the libraries folder exists.

When i imagined using the manifest, my thought process was that if a manifest exists, the server must've ran before, in turn a `libraries` folder should exist.

However this is not reliable as `SETUP_ONLY` can be passed, creating the jar and manifest, but no libraries.

itzg (approved):
Thanks for fixing this

itzg:
Published as https://github.com/itzg/mc-image-helper/releases/tag/1.57.3

PR #765, opened 2026-05-11
build(deps): Fix Junit and Netty dependency errors in Java8 build

harrisonablack:
Currently test fails with message; `The automatic loading of test framework implementation dependencies has been deprecated. Solution: Declare the desired test framework directly on the test suite or explicitly declare the test framework implementation dependencies on the test's runtime classpath`

Gradle has deprecated auto-loading of test launchers and now must manually be declared

["If using JUnit 5, an explicit runtimeOnly dependency on junit-platform-launcher is required in addition to the existing implementation dependency on the test engine."](https://docs.gradle.org/8.12/userguide/upgrading_version_8.html#test_framework_implementation_dependencies)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -100,9 +100,11 @@ dependencies {
     // http3/quic is not functional on Alpine due to absence of glibc/ld-linux
     // Error loading shared library ld-linux-x86-64.so.2:
     //    No such file or directory (needed by /tmp/libnetty_quiche42_linux_x86_641323735684306068308.so)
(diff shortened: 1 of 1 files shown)


harrisonablack:
Tests where also all failing on `Caused by: java.lang.ExceptionInInitializerError at Http3.java:35`. This was fixed in #682 however not brought back into java 8 branch.

Added commit to backport into java8 branch

itzg (approved):
Thanks for prepping the java8 branch.

PR #767, opened 2026-05-11
feat: add install-oci-pack command

ChipWolf:
Adds an `install-oci-pack` subcommand that pulls OCI artifacts from any distribution-spec registry and writes layer blobs to disk in manifest order, for the `GENERIC_PACK[S]` handler in docker-minecraft-server.

The [ORAS Java SDK](https://github.com/oras-project/oras-java) requires Java 17+ and this project targets Java 8, so this is a minimal client instead covering the distribution-spec surface needed for artifact pulls: bearer challenge/response auth with token caching, manual redirect handling, and streaming SHA-256 verification.

Reads credentials from `~/.config/containers/auth.json` or `~/.docker/config.json`. No credHelper support.

The output directory acts as a content-addressed cache, so shared layers across sibling artifacts are only downloaded once.

Ref: https://github.com/itzg/docker-minecraft-server/issues/4038

### DIFF
--- a/src/main/java/me/itzg/helpers/McImageHelper.java
+++ b/src/main/java/me/itzg/helpers/McImageHelper.java
@@ -32,6 +32,7 @@ import me.itzg.helpers.modrinth.InstallModrinthModpackCommand;
 import me.itzg.helpers.modrinth.ModrinthCommand;
 import me.itzg.helpers.modrinth.VersionFromModrinthProjectsCommand;
 import me.itzg.helpers.mvn.MavenDownloadCommand;
+import me.itzg.helpers.oci.InstallOciPackCommand;
 import me.itzg.helpers.paper.InstallPaperCommand;
 import me.itzg.helpers.patch.PatchCommand;
 import me.itzg.helpers.properties.SetPropertiesCommand;
@@ -78,6 +79,7 @@ import picocli.CommandLine.Spec;
         InstallForgeCommand.class,
         InstallModrinthModpackCommand.class,
         InstallNeoForgeCommand.class,
+        InstallOciPackCommand.class,
         InstallPaperCommand.class,
         InstallPurpurCommand.class,
         InstallQuiltCommand.class,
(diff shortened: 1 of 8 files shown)


itzg:
Thanks...but, I wish we'd had some design talk first. 

[Java 11 is actually the target version](https://github.com/itzg/mc-image-helper/blob/794ec60471a0db32ee27c9ede74dc8d4626c2b49/build.gradle#L55), but I would have been ok bumping that to 17 and relegating both java8 and java11 images to use the https://github.com/itzg/mc-image-helper/tree/java8 maintenance branch releases. By not using the ORAS SDK, the good news is this PR can probably be cherry picked over to that branch. This and [a recent enhancement](https://github.com/itzg/docker-minecraft-server/pull/4046#issuecomment-4415539143) start to make me wonder if I should branch off java8 (and 11) and put those into maintenance mode. 

I have been gradually getting rid of Apache HC5 and have put a lot of focus into ensuring new code only uses the Netty Reactor wrapper that starts at https://github.com/itzg/mc-image-helper/blob/master/src/main/java/me/itzg/helpers/http/Fetch.java . It would be great if you could switch the Oras client to use that; however, I know that's not a trivial ask since Fetch is all reactive. 

I'll give the PR a proper review when I can sit at a computer for a while.

ChipWolf:
I have another branch using the ORAS SDK, I can relegate <17 to that java8 branch and open a new PR using the SDK if you'd prefer, otherwise I'll just pick the hand-rolled client across @itzg

itzg:
Can you point me to that branch? I'm curious what that approach looks like. It would indirectly address my Apache HC5 concern also.

ChipWolf:
@itzg mentioned above

itzg:
> @itzg mentioned above
> 
> 

Sorry, I must have been looking at an un-refreshed page.

itzg:
I take back my apology 😄 , your comment with the info about your branch is no longer there

image

ChipWolf:
I removed a comment here b/c it's all in the other PR description (:

vivaldi_tNfJHYrwhK

ChipWolf:
Duplicate pull request #768

PR #768, opened 2026-05-12
Add install-oci-pack command (ORAS SDK)

ChipWolf:
Adds an `install-oci-pack` subcommand that pulls OCI artifacts from any distribution-spec registry and writes layer blobs to disk in manifest order, for the `GENERIC_PACK[S]` handler in docker-minecraft-server.

Alternative implementation to #767, using the [`land.oras:oras-java-sdk`](https://github.com/oras-project/oras-java) instead of a hand-rolled client.

The SDK uses `java.net.http.HttpClient` (the JDK built-in) and has no HC5 dependency, so this change adds no new `org.apache.hc.*` imports. It targets Java 17 bytecode, so the toolchain bumps from 11 to 17. About 150 fewer LOC than #767, since the SDK handles manifest parsing, bearer auth, redirects, and digest verification.

The command surface (`--ref`, `--output-directory`, `--auth-file`, `--filename-strategy`, `--layer-list-file`) and content-addressed caching behaviour match #767.

For itzg/docker-minecraft-server#4038.

### DIFF
--- a/src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
+++ b/src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
@@ -114,6 +114,7 @@ public class InstallOciPackCommand implements Callable<Integer> {
                         throw new GenericException(
                             "Failed to download OCI layer " + digest + ": " + e.getMessage(), e);
                     }
+                    verifyDownloaded(target, digest);
                 }
                 written.add(target);
             }
@@ -214,19 +215,56 @@ public class InstallOciPackCommand implements Callable<Integer> {
             return false;
         }
         try {
-            final MessageDigest sha = MessageDigest.getInstance("SHA-256");
-            try (InputStream in = Files.newInputStream(target)) {
-                final byte[] buf = new byte[8192];
-                int n;
-                while ((n = in.read(buf)) > 0) {
-                    sha.update(buf, 0, n);
-                }
-            }
-            return ("sha256:" + toHex(sha.digest())).equalsIgnoreCase(expectedDigest);
-        } catch (IOException | NoSuchAlgorithmException e) {
+            return computeSha256(target).equalsIgnoreCase(expectedDigest);
+        } catch (IOException e) {
             log.debug("Could not verify existing layer at {}: {}", target, e.getMessage());
             return false;
         }
+    }
+
+    // The ORAS SDK only verifies blob bytes against the Docker-Content-Digest
+    // response header, and silently skips verification when that header is
+    // absent (the OCI distribution spec marks it SHOULD, not MUST). Recompute
+    // the digest ourselves against the manifest's declared value so a
+    // misbehaving registry or stripping proxy cannot serve us tampered bytes.
+    private static void verifyDownloaded(Path target, String expectedDigest) {
+        if (!"sha256:".regionMatches(true, 0, expectedDigest, 0, "sha256:".length())) {
+            return;
+        }
+        final String actual;
+        try {
+            actual = computeSha256(target);
+        } catch (IOException e) {
+            throw new GenericException(
+                "Failed to verify digest of " + target + ": " + e.getMessage(), e);
+        }
+        if (!actual.equalsIgnoreCase(expectedDigest)) {
+            try {
+                Files.deleteIfExists(target);
+            } catch (IOException ignored) {
+                // best-effort cleanup
+            }
+            throw new GenericException(String.format(
+                "Digest mismatch for OCI blob: expected %s, got %s",
+                expectedDigest, actual));
+        }
+    }
+
+    private static String computeSha256(Path target) throws IOException {
+        final MessageDigest sha;
+        try {
+            sha = MessageDigest.getInstance("SHA-256");
+        } catch (NoSuchAlgorithmException e) {
+            throw new AssertionError("SHA-256 unavailable in this JRE", e);
+        }
+        try (InputStream in = Files.newInputStream(target)) {
+            final byte[] buf = new byte[8192];
+            int n;
+            while ((n = in.read(buf)) > 0) {
+                sha.update(buf, 0, n);
+            }
+        }
+        return "sha256:" + toHex(sha.digest());
     }
 
     private static String toHex(byte[] bytes) {
--- a/src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
+++ b/src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
@@ -6,7 +6,6 @@ import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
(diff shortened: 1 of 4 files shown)


itzg:
I’ll also look closer at this one, but so far I think I prefer it. 

I’m now inclined to put the java 8 - 16 itzg/mjnecraft-server tagged images into a maintenance mode with no new features and as such keep the “java8” branch here in mc-image-helper in a non-feature maintenance mode. 

…however, are you needing to run Minecraft versions like 1.12.2, 1.16.5, etc? No point in limiting the enhanced image variants if it defeats your use cases.

ChipWolf:
I only run >1.20 personally. Thanks!

src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
```diff
@@ -0,0 +1,235 @@
...
+
+    @Option(names = "--layer-list-file", paramLabel = "FILE",
+        description = "Write each pulled layer's absolute path on its own line to this file,"
+            + " in manifest layer order. Suitable for `mapfile -t ... < FILE` in shell."
+            + " When omitted, layer paths are also printed to stdout for interactive use.")
+    Path layerListFile;
+
+    // Allows tests to capture stdout without intercepting the JVM's
+    @Setter
+    private PrintStream out = System.out;
```
itzg: This is used in other unit tests

https://github.com/stefanbirkner/system-lambda#systemout-and-systemerr

src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
```diff
@@ -0,0 +1,235 @@
...
+
+    // Visible for tests; production code always uses HTTPS
+    @Setter
```
itzg: In that case


```suggestion
    // Visible for tests; production code always uses HTTPS
    @Setter(AccessLevel.PACKAGE)
```

src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
```diff
@@ -0,0 +1,235 @@
...
+    String scheme = "https";
+
+    public enum FilenameStrategy {
+        title,
+        digest
```
itzg, line 78: It would be handy for later maintenance to give each of these brief javadoc

src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
```diff
@@ -0,0 +1,235 @@
...
+
+    // Allows tests to capture stdout without intercepting the JVM's
+    @Setter
+    private PrintStream out = System.out;
+
+    // Visible for tests; production code always uses HTTPS
+    @Setter
+    String scheme = "https";
+
+    public enum FilenameStrategy {
```
itzg, line 71: It would be handy to give each of these a javadoc

src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
```diff
@@ -0,0 +1,235 @@
...
+            + " When omitted, layer paths are also printed to stdout for interactive use.")
+    Path layerListFile;
+
+    // Allows tests to capture stdout without intercepting the JVM's
+    @Setter
+    private PrintStream out = System.out;
+
+    // Visible for tests; production code always uses HTTPS
+    @Setter
+    String scheme = "https";
```
itzg: ...now that I see the usage, this seems like an indirect way to indicate security registry access or not. I would rather it be a boolean `insecure` or even better derived from the `ref` scheme; however, I can see that "oci://" would be ambiguous.

src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+
+    private String filenameFor(Layer layer) {
+        final String candidate;
+        if (filenameStrategy == FilenameStrategy.digest) {
+            candidate = layer.getDigest();
+        } else {
+            final String title = layerTitle(layer);
+            candidate = (title != null && !title.isEmpty()) ? title : layer.getDigest();
```
itzg: Can be simplified since this is Java 11+ code


```suggestion
            candidate = !String.isBlank(title) ? title : layer.getDigest();
```

itzg (commented):
Yeah, let's go with version. This looks awesome. 

I'll start thinking about where/how to have the image variants less than java17 use only the maintenance mode releases of mc-image-helper. I'm thinking it'll be a separate branch over there like the java8 branch here.

itzg:
FYI I have added a new subcommand

https://github.com/itzg/mc-image-helper#has-feature

that will allow for a single mainline branch in https://github.com/itzg/docker-minecraft-server but allow the scripts to query if a feature is available, such as the new subcommand in this PR. For the Java 8 - 16 image variants, using the https://github.com/itzg/mc-image-helper/releases/tag/1.51.3-java8 type releases, that query will return exit code 1 since we won't backport things that required Java 17. I'm thinking the script call point can gracefully fail the startup indicating that the requested feature isn't available for that image java variant.

Does seem appropriate to you? If so, I'll capture all this in the README here and https://github.com/itzg/docker-minecraft-server/blob/master/docs/misc/contributing/development.md

src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
ChipWolf: `out` field replaced with `tapSystemOutNormalized` capture (matches `IniPathCommandTest`)

src/main/java/me/itzg/helpers/oci/InstallOciPackCommand.java
ChipWolf: `scheme` String swapped for an `insecure` boolean with `@Setter(AccessLevel.PACKAGE)`

ChipWolf:
@itzg this one is potentially in a good place now for re-review, tyvm

itzg:
Sorry, for being slow getting back to this. It's been a busy couple of weeks at home.

ChipWolf:
> Sorry, for being slow getting back to this. It's been a busy couple of weeks at home.

no rush! hope all is well

itzg:
Yep. Wife's birthday, anniversary, house renovation. All good stuff 😁

itzg (approved):
Thanks!

itzg:
Released in https://github.com/itzg/mc-image-helper/releases/tag/1.59.0

Next can update itzg/minecraft-server such as pinning all the java16 - java8 here

https://github.com/itzg/docker-minecraft-server/blob/58781c37246f566d0a53472056accdd8ac4d1eb7/.github/workflows/build.yml#L73-L95

and then the script changes, of course, wrapped in the new `has-feature`.

itzg:
Actually, I need to go ahead and pin those to get the builds back to happy

https://github.com/itzg/docker-minecraft-server/pull/4077

PR #776, opened 2026-05-26
neoforge: fix version extraction from version.json

EZForever:
NeoForge installer's version is formatted like `neoforge-21.1.231`, i.e. only two parts. The current logic thus always fails to parse this, and throws an `Unexpected format of id from Forge installer's version.json` exception as long as NeoForge is being installed from a local installer. I mean, *c'mon*, someone got to have discovered this in the past 3 months...

PR #778, opened 2026-05-27
neoforge: handle beta local installers

EZForever:
This is a follow-up to my last fix, #776. *Beta* versions of NeoForge do have a three-part `id` field, like `neoforge-26.1.2.66-beta`. My last fix did not consider this, and will fail the parsing logic again if a beta local installer is used. Sorry for the additional noise.


itzg:
Now it's time for a unit test. I started looking at that last night and was thinking it should include a 26.* beta and the one you used. The 26 stuff broke the download metadata version parsing. 

The test class is already there. Just needs a parameterized test method for the parsing method you updated.

PR #789, opened 2026-06-09
ServerLibraryCleaner: Add Forge support for server library cleaning

harrisonablack:
When forge installs, it creates a "*-shim.jar" file, this jar is the entrypoint for the server, and contains the `bootstrap-shim.list`.

The `bootstrap-shim.list` is then read like other Jar types, parsed, and the diff between the installed and required libs are deleted.

## Backwards Compatability
This way of reading libraries is only compatible and tested with Forge 49.0.2 installer for Minecraft 1.20.3 or newer.

I believe @itzg in a previous comment you mentioned preference for a best-effort coverage, and not to be concerned with full backwards compatibility, however if you'd prefer support for prior versions, i'm happy to look at it

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,6 +542,7 @@ loader instances
 
 ```
 Usage: mc-image-helper install-forge [-h] [--force-reinstall]
+                                     [--clean-libraries]
                                      [--minecraft-version=VERSION]
                                      [--output-directory=DIR]
                                      [--results-file=FILE]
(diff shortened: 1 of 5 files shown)


itzg:
Haven't reviewed yet, but yes I'm still good with best effort support for modern versions. Can always add more heuristics later for older/others.

itzg (approved):
Thanks!

itzg:
Released in https://github.com/itzg/mc-image-helper/releases/tag/1.61.0

PR #812, opened 2026-07-15
Vanilla in image helper

jcotton42:
This is a prerequisite for itzg/docker-minecraft-server#1818, and moves the vanilla Minecraft resolution and installation to mc-image-helper so it can be used by the packwiz installer.


jcotton42:
I still want to add tests, but need to learn how Wiremock works before I can do so.

jcotton42:
I think I figured out Wiremock last night, so I'll add tests.

src/main/java/me/itzg/helpers/vanilla/VanillaInstaller.java
```diff
@@ -0,0 +1,110 @@
...
+
+        Manifests.save(outputDirectory, VanillaManifest.ID, newManifest);
+
+        log.debug("Populating results file {}", resultsFile);
+
+        try (ResultsFileWriter writer = new ResultsFileWriter(resultsFile)) {
+            writer.writeType("VANILLA");
+            writer.writeServer(outputDirectory.resolve(newManifest.getServerEntry()));
+            writer.writeVersion(newManifest.getMinecraftVersion());
+        }
```
harrisonablack: I would wrap the Results File writing logic in a null check

If you don't supply a Results File, mc-image-helper will successfully install the vanilla server, however throw an error as its trying to read a null `Path`

```suggestion
    if (resultsFile != null) {
        try (ResultsFileWriter writer = new ResultsFileWriter(resultsFile)) {
            writer.writeType("VANILLA");
            writer.writeServer(outputDirectory.resolve(newManifest.getServerEntry()));
            writer.writeVersion(newManifest.getMinecraftVersion());
        }
     }
```

src/main/java/me/itzg/helpers/vanilla/VanillaInstaller.java
```diff
@@ -0,0 +1,110 @@
...
+    private final SharedFetch sharedFetch;
+    private final MinecraftVersionsApi versionsApi;
+
+    public VanillaInstaller(SharedFetch sharedFetch) {
+        this.sharedFetch = sharedFetch;
+        this.versionsApi = new MinecraftVersionsApi(sharedFetch);
+    }
+
+    public void install(String version, Path outputDirectory, Path resultsFile, boolean forceReinstall) throws IOException {
+        final VanillaManifest prevManifest = Manifests.load(outputDirectory, VanillaManifest.ID, VanillaManifest.class);
```
harrisonablack, line 31: Theres no validation on the outputDirectory.

If you pass a directory that doesn't exist, it produces an obfuscated `NoSuchFileException`. I would add a brief check before you Load the manifest, checking if the directory exists, either creating the directory or exiting with an error

```suggestion
        if (Files.isDirectory(Path.of(...))) {}

        final VanillaManifest prevManifest = Manifests.load(outputDirectory, VanillaManifest.ID, VanillaManifest.class);
```

itzg: This is actually bit of a Java-ism since exceptions are not just annoying, they're powerful. In this case, why bother preemptively checking if the directory exists if the final file operation at the OS level is always going to be checking and throw an `IOException`? Catching and wrapping exceptions is the way to de-obfuscate an exception, much like `fmt.Errorf("... %w")` in Go.

Now if the operation were to break things by creating a file when the path actually should have been a directory, then a preemptive conditional makes sense.

itzg (commented):
Looks great with @harrisonablack's suggested results file change. The tests will be even better 😄 ..and yes, Wiremock has bit of a learning curve. Worth it once you're there.

jcotton42:
Pushed some tests for MinecraftVersionsApi, mostly as a way to make sure I understand WireMock. Next time I touch this I'll do the actual tests for the vanilla installer, and then it should be ready for merge.

P.S. all the jars are fake, really just zip files with a single text file entry to ensure all the JARs have different SHAs.

src/test/resources/versions/__files/1.6.json
```diff
@@ -0,0 +1,9 @@
+{
+  "downloads": {
+    "server": {
+      "sha1": "7cafd8992ed3ed977bfbaf6b35b0b1ddea01f400",
+      "size": 166,
+      "url": "{{request.baseUrl}}/jars/1.6.jar"
```
jcotton42, line 6: This doesn't actually match the pathing Mojang uses, but since the official version API hands out absolute URLs, I figured it didn't really matter for our test data?

itzg: Yeah this is where the wiremock'ing can get real tedious since you have to stub out referenced download URLs like that.

itzg:
I just remembered that I still need to upgrade Wiremock now that the mainline branch is no longer stuck on Java 8. Some of the template stub stuff might work better in newer versions, so let me know if you're fighting it too much and I'll slide in that upgrade.

jcotton42:
It was no big deal, the only snag I hit was 2.x doesn't let you put multiple mappings in one JSON, but that was easily split up with a bit of PowerShell.

One aside though, does SharedFetch have a good way to compute hashes during downloads? Like inherently or by somehow sticking something into the downloading? I'd like to verify the downloaded JARs against the SHA in the manifest.

I could read the just-downloaded JAR off disk, but that feels inelegant 😄

Edit: looking at the wiki reference for version_manifest_v2.json, each version also has a sha1 field that is the hash of the pointed-to version manifest. That might also be nice to verify before deserializing the JSON.

Edit2: for checking the jar, looks like I could use a `FileDownloadedHandler`, not sure about the json manifest though.

itzg:
There's this pattern but it's the way you don't like 😄 

https://github.com/itzg/mc-image-helper/blob/f6f4965eb990a2426bb70280cebf981ff7dcb07a/src/main/java/me/itzg/helpers/curseforge/FileHashVerifier.java#L42

itzg:
...the one think I don't like about Reactor Netty is that it's reactive 😉 What I actually mean by that is that things that should be simple like calculating the checksum as the bytes come through should be easy, but is weirdly awkward with reactive since the processing pipeline is stateless.

jcotton42:
That's unfortunate. I'll add checksumming of the jar, and for now we'll just have to rely on HTTPS for ensuring the manifest is valid.

src/main/java/me/itzg/helpers/files/FileHashInvalidException.java
jcotton42, line 1: I took the liberty of moving this to `files` from `curseforge` since it's now being used across modules and was generic.

jcotton42:
I think that should be good as far as tests are concerned, unless you can think of other cases.

jcotton42:
I just realized, since the jars are never actually executed, they don't even need to be zips. They could just be text files, which will be nicer on git. One second.

itzg:
> I just realized, since the jars are never actually executed, they don't even need to be zips. They could just be text files, which will be nicer on git. One second.

I knew there was something I wanted to tell you before with regard to wiremock'ing. Yeah I always just have the "jars" just be a line of text to verify.

src/main/java/me/itzg/helpers/vanilla/InstallVanillaCommand.java
```diff
@@ -0,0 +1,47 @@
...
+    @SuppressWarnings("unused")
+    @Option(names = {"--help", "-h"}, usageHelp = true)
+    boolean help;
+
+    @Option(names = "--output-directory", defaultValue = ".", paramLabel = "DIR")
+    Path outputDirectory;
+
+    @Option(names = "--results-file", description =
+        "A key=value file suitable for scripted environment variables. Currently includes"
+            + "\n  SERVER: the entry point jar or script", paramLabel = "FILE")
```
itzg: Thanks for documenting the contents. Also will include the resolved `VERSION`

src/main/java/me/itzg/helpers/vanilla/VanillaInstaller.java
```diff
@@ -0,0 +1,132 @@
...
+                .block();
+        }
+
+        if (prevManifest != newManifest) {
+            Manifests.cleanup(outputDirectory, prevManifest, newManifest, log);
+        }
+
+        Manifests.save(outputDirectory, VanillaManifest.ID, newManifest);
+
+        log.debug("Populating results file {}", resultsFile);
```
itzg: Move into if-block

itzg (changes requested):
This look great! Just a couple of super minor tweaks.

itzg (approved):

jcotton42:
I've taken itzg/docker-minecraft-server#4165 out of draft so it can be looked at once this is merged.

itzg:
> I've taken [itzg/docker-minecraft-server#4165](https://github.com/itzg/docker-minecraft-server/pull/4165) out of draft so it can be looked at once this is merged.

I'll cut a release and let Renovate apply that

Issue #810, opened 2026-07-14
Add artifact download for github


SgtMate:
**Idea:** 
Add the possibility to download specific (build) artifacts from github.

**Use-case:** 
Enable the download of specific builds/releases of a github project. I would like to use this to handle the download of development versions of GTNH for the minecraft server project.

**Curl Example:**
```bash
curl -L \
  --output test.zip \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer <TOKEN>" \
  -H "X-GitHub-Api-Version: 2026-03-10" \
  https://api.github.com/repos/GTNewHorizons/DreamAssemblerXXL/actions/artifacts/8131458572/zip
```

itzg:
@SgtMate , what are you thinking would be the inputs to a new subcommand for that?

Perhaps:
- org/repo
- release name
  - \+ regex to match desired artifact
- or just artifact ID, I'm guessing from your curl example

Looking at https://github.com/GTNewHorizons/DreamAssemblerXXL I don't even see a releases section, so I'm not sure how to relate that to your example curl

SgtMate:
@itzg 
For my usecase org/repo and artifact ID is most suitable as the build artifacts aren't necessarily part of a release.
Alternatively [the run ID of the workflow](https://docs.github.com/en/rest/actions/artifacts?apiVersion=2026-03-10&versionId=free-pro-team%40latest&productId=rest#list-workflow-run-artifacts) could be used and combined with a regex for matching. 

The DreamAssembler does not have releases, there are daily builds of the modpack, [here](https://github.com/GTNewHorizons/DreamAssemblerXXL/actions). Final releases are published on the website (already implemented). Now I want to add support for people who want to try the daily/dev releases by hand eg. for debugging the modpack or just testing new features.

itzg:
Thanks for clarifying. Now I see the **build** qualifier in your original feature description.

Did you have interest in implementing this? Just wanted to check before I jump on it or maybe @harrisonablack does 😄

SgtMate:
@itzg feel free to implement this. Thanks 😊

SgtMate:
@itzg one point i might want to add is that github also provides checksums for artifacts. Maybe they could be checked after download?

itzg:
Yep, can try to add that too. I already have checksum logic used for the CurseForge downloads.

harrisonablack:
> Thanks for clarifying. Now I see the **build** qualifier in your original feature description.
> 
[...]

PR #816, opened 2026-07-16
feat: Add Github action artifact download support

harrisonablack:
Add sub-command under `GithubCommands.java` to allow downloading of specific Github action artifacts

refs: #810

### DIFF
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -125,7 +125,7 @@ public class DownloadArtifactCommand implements Callable<Integer> {
             if (unzip) {
                 log.info("Unzipping artifact");
                 unzipArtifact(download, outputDirectory, overwrite);
-                Files.deleteIfExists(download);
+                Files.delete(download);
             }
         }
 
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -201,7 +201,7 @@ public class DownloadArtifactCommand implements Callable<Integer> {
                 }
             }
         } catch (IOException e) {
-            throw new RuntimeException("Failed to extract artifact archive", e);
+            throw new GenericException("Failed to extract artifact archive", e);
         }
     }
 
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -62,8 +62,8 @@ public class DownloadArtifactCommand implements Callable<Integer> {
         description = "Regular expression that must match exactly one artifact")
     private Pattern artifactPattern;
 
-    @Option(names = "--print-artifact", description = "Prints artifact name", defaultValue = "false")
-    private Boolean printArtifact;
+    @Option(names = "--output-filename", description = "Prints artifact name", defaultValue = "false")
+    private Boolean outputFilename;
 
     @Option(names = "--no-download", description = "Doesn't download artifact", defaultValue = "false")
     private Boolean noDownload;
@@ -101,8 +101,8 @@ public class DownloadArtifactCommand implements Callable<Integer> {
             Mono<Artifact> candidate = resolveArtifact(client)
                 .switchIfEmpty(Mono.error(new GenericException("Github client failed to find an artifact")));
 
-            if (printArtifact) {
-                log.info("Resolved artifact {}", candidate.block().getName());
+            if (outputFilename) {
+                System.out.println(candidate.block().getName());
             }
 
             if (noDownload) {
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -101,7 +101,10 @@ public class DownloadArtifactCommand implements Callable<Integer> {
 
             Mono<Artifact> candidate = resolveArtifact(client)
                 .switchIfEmpty(Mono.error(new GenericException("Github client failed to find an artifact")))
-                .doOnNext((artifact) -> System.out.println(artifact.getName()));
+                .doOnNext((artifact) -> {
+                    if (outputFilename) {
+                        System.out.println(artifact.getName());
+                    }});
 
             if (noDownload) {
                 candidate.block();
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -119,7 +119,8 @@ public class DownloadArtifactCommand implements Callable<Integer> {
                 .doOnNext(artifact -> {
                     log.info("Downloading artifact {}", artifact.getName());
                 }) 
-                .flatMap(artifact -> downloadArtifact(client, artifact)).block();
+                .flatMap(artifact -> downloadArtifact(client, artifact))
+                .block();
 
 
             if (unzip) {
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -104,6 +104,7 @@ public class DownloadArtifactCommand implements Callable<Integer> {
                 .doOnNext((artifact) -> System.out.println(artifact.getName()));
 
             if (noDownload) {
+                candidate.block();
                 return ExitCode.OK;
             }
 
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -24,7 +24,7 @@ import picocli.CommandLine.ParentCommand;
 import reactor.core.publisher.Mono;
 
 @Command(name = "download-artifact",
-    description = "Download an artifact from a successful GitHub Actions workflow as received or unzipped")
+    description = "Download an artifact from a successful GitHub Actions workflow")
 @Slf4j
 public class DownloadArtifactCommand implements Callable<Integer> {
 
@@ -37,25 +37,25 @@ public class DownloadArtifactCommand implements Callable<Integer> {
     static class RunSelector {
 
         @Option(names = "--workflow", paramLabel = "ID|FILE",
-            description = "Workflow ID or filename whose latest successful run should be used")
+            description = "Query this workflow for the latest successful run to download an artifact")
         private String workflow;
 
         @Option(names = "--run-id", paramLabel = "ID",
-            description = "Specific workflow run ID to use")
+            description = "Query this specific workflow run to download an artifact")
         private Long runId;
     }
 
     @ArgGroup(multiplicity = "1")
     private RunSelector runSelector;
 
-    @Option(names = {"--output-directory", "-o"}, defaultValue = ".")
+    @Option(names = {"--output-directory", "-o"}, defaultValue = ".", description = "Output directory of downloaded artifact")
     private Path outputDirectory;
 
-    @Option(names = "--unzip", description = "Extract the downloaded response and remove it afterward")
+    @Option(names = "--unzip", description = "Extract the downloaded artifact")
     private boolean unzip;
 
     @Option(names = "--overwrite", defaultValue = "false",
-        description = "When extracting zip, overwrite existing files")
+        description = "Overwrite existing files when extracing zip")
     private boolean overwrite;
 
     @Option(names = "--name-pattern", required = true,
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -87,13 +87,6 @@ public class DownloadArtifactCommand implements Callable<Integer> {
     @Override
     public Integer call() throws Exception {
 
-        // Github API requires a token with "'Actions' repository
-        // permissions (read)" to download an artifact
-        // https://docs.github.com/en/rest/actions/artifacts?apiVersion=2026-03-10#download-an-artifact
-        if (parent.token == null) {
-            throw new IllegalArgumentException("Must provide a github token to query artifact data");
-        }
-
         if (!Files.isDirectory(outputDirectory)) {
             try {
                 Files.createDirectory(outputDirectory);
@@ -114,6 +107,13 @@ public class DownloadArtifactCommand implements Callable<Integer> {
 
             if (noDownload) {
                 return ExitCode.OK;
+            }
+
+            // Github API requires a token with "'Actions' repository
+            // permissions (read)" to download an artifact
+            // https://docs.github.com/en/rest/actions/artifacts?apiVersion=2026-03-10#download-an-artifact
+            if (parent.token == null) {
+                throw new IllegalArgumentException("Must provide a github token to query artifact data");
             }
 
             Path download = candidate
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -88,10 +88,12 @@ public class DownloadArtifactCommand implements Callable<Integer> {
             throw new IllegalArgumentException("Must provide a github token to query artifact data");
         }
 
-        try {
-            Files.createDirectory(outputDirectory);
-        } catch (IOException e) {
-            log.error("Failed to create output directory {}", outputDirectory.toAbsolutePath().toString());
+        if (!Files.isDirectory(outputDirectory)) {
+            try {
+                Files.createDirectory(outputDirectory);
+            } catch (IOException e) {
+                log.error("Failed to create output directory {}", outputDirectory.toAbsolutePath().toString());
+            }
         }
 
         try (SharedFetch sharedFetch = Fetch.sharedFetch("github download-artifact", sharedFetchArgs.options())) {
--- a/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
+++ b/src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
@@ -59,7 +59,7 @@ public class DownloadArtifactCommand implements Callable<Integer> {
     private boolean overwrite;
 
     @Option(names = "--artifact-pattern", required = true,
-        description = "Regular expression that must match exactly one artifact name")
+        description = "Regular expression that must match exactly one artifact")
     private Pattern artifactPattern;
 
     @ParentCommand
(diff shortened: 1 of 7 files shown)


harrisonablack:
@itzg @SgtMate 

You can either pass a specific workflow run id, or a workflow name, along with organisation/repository.

Command then queries actions, selecting either the latest succesful action for a specific workflow, or the already provided workflow id. Then pattern matches against artifacts and downloads the matching artifact.

```bash
Usage: mc-image-helper github download-artifact [-h] [--overwrite] [--unzip]
       --artifact-pattern=<artifactPattern> [-o=<outputDirectory>]
       (--workflow=ID|FILE | --run-id=ID) [[--use-http2] [--wiretap]] org/repo
Download an artifact from a successful GitHub Actions workflow
      org/repo
      --artifact-pattern=<artifactPattern>
                           Regular expression that must match exactly one
                             artifact name
  -h, --help
  -o, --output-directory=<outputDirectory>
                           Output directory of downloaded artifact
      --overwrite          Overwrite existing files when extracing zip
      --run-id=ID          Query this specific workflow run to download an
                             artifact
      --unzip              Extract the downloaded artifact
      --use-http2          Whether to use HTTP/2.
                           Default: true
                           Env: FETCH_USE_HTTP2
      --wiretap            Whether to enable Reactor Netty wiretap logging.
                             Make sure to set logging level to trace.
                           Default: false
                           Env: FETCH_WIRETAP
      --workflow=ID|FILE   Query this workflow for the latest successful run to
                             download an artifact
```

For example, with @SgtMate use case, the below queries the latest successful daily build, downloading the Java17 server arifact, and extracting it to ./gtnh

```bash
GH_TOKEN="$(gh auth token)" ./gradlew run --args="github download-artifact --workflow daily-modpack-build.yml --name-pattern='GTNH-daily-\d{4}-\d{2}-\d{2}\+\d+-server-java17-25\.zip' -o=./gtnh --unzip GTNewHorizons/DreamAssemblerXXL"
```

> Github requires a token to be passed when downloading Artifacts 
https://docs.github.com/en/rest/actions/artifacts?apiVersion=2026-03-10

SgtMate:
this looks great! this is exactly what i need. one thing i want to add tho is the `--output-filename` option (from the get command)

harrisonablack:
> this looks great! this is exactly what i need. one thing i want to add tho is the `--output-filename` option (from the get command)

When you want to change output name, do you want the file still as a zip i assume? And the zip renamed? Or whatever format the Artifact is

SgtMate:
So what I need is for the command to return the filename that was downloaded. (independent of if its unpacked or not).
`--output-filename     Output the resulting filename`

itzg:
This looks great @harrisonablack and I like @SgtMate 's request.

SgtMate:
another question, just so i understand. is it correct, that the gh token is not handed to the command directly? so i need to set it in env?

itzg:
The token can be passed either way:

```
    @Option(names = "--token", defaultValue = "${env:GH_TOKEN}",
        description = "An access token for GitHub to elevate rate limit vs anonymous access"
    )
```

however, within itzg/minecraft-server the expectation is that the user has set the container env var to allow for token access across the whole startup.

SgtMate:
@itzg ok get it, then i will adhere to the existing processes.

itzg:
Wait, that's bit of a usability problem. Users installing GTNH aren't expecting to need a GitHub token. 

@harrisonablack are you certain the APIs used need authenticated access?

SgtMate:
@itzg thats fine, I'll cleary state this in the addition to the documentation.

harrisonablack:
> Wait, that's bit of a usability problem. Users installing GTNH aren't expecting to need a GitHub token.
> 
> @harrisonablack are you certain the APIs used need authenticated access?

Yes, when i remove the null token check, and don't set env, command fails with a 401.

I believe its only the Artifact download endpoint that requires auth

```txt
[mc-image-helper] 23:18:25.773 ERROR : 'download-artifact' command failed. Version is ???: FailedRequestException: HTTP request of https://api.github.com/repos/GTNewHorizons/DreamAssemblerXXL/actions/artifacts/8398696458/zip failed with 401 Unauthorized: Trying to retrieve file
```

```bash
./gradlew run --args="github download-artifact --workflow daily-modpack-build.yml --artifact-pattern='GTNH-daily-\d{4}-\d{2}-\d{2}\+\d+-server-java17-25\.zip' -o=./gtnh --unzip GTNewHorizons/DreamAssemblerXXL"
```

harrisonablack:
https://docs.github.com/en/rest/actions/artifacts?apiVersion=2026-03-10#download-an-artifact

> The fine-grained token must have the following permission set:
> - 'Actions' repository permissions (read)

SgtMate:
another thing i need to add to my request is some kind of dry-run functionality. where only the resulting download candidate is returned without downloading it. I need this to check if an update is even required once the container is restarted. Currently im just comparing the old and new filename. This would greatly help in reducing unneeded downloads. 

@harrisonablack you think you can implement this with githubs given api?

itzg:
It's almost like it needs a "skip existing" like the `get` command, but with a 3-way outcome

- New file was available, here's its name
- Existing file was already present
- Something bad happened

harrisonablack:
> another thing i need to add to my request is some kind of dry-run functionality. where only the resulting download candidate is returned without downloading it. I need this to check if an update is even required once the container is restarted. Currently im just comparing the old and new filename. This would greatly help in reducing unneeded downloads.
> 
> @harrisonablack you think you can implement this with githubs given api?

@SgtMate My thought process is yes, but what are we trying to do? What level of automation of the process do we want to reach?

If you want to automate updating / installing daily builds of GTNH, i would lean towards a manifest similar to Paper, where when the server is installed, a file is created that states which version is installed.

Then when we go to restart the container, we can check for a new build and can compare artifact names / dates before downloading.

If we then automate isntalling / updating, my next questions are then GTNH specific; which files are safe to replace when updating a server, and which files do we need to keep? I'm not overly familiar with Modpacks / Modpack servers.

I would lean more towards implementing a `install-gtnh` command in mc-image-helper to automate this process.

But i digress, yes, i can add another function / command argument to return the name of the artifact only

harrisonablack:
> It's almost like it needs a "skip existing" like the `get` command, but with a 3-way outcome
> 
> * New file was available, here's its name
> * Existing file was already present
> * Something bad happened

I did add an `--overwrite` flag, which when set overwrites already existing files during the unzip. Default behaviour is to skip existing files

itzg:
> I would lean more towards implementing a `install-gtnh` command in mc-image-helper to automate this process.

I gently suggested that too 😉 

https://github.com/itzg/docker-minecraft-server/pull/4141#issuecomment-4849279649

Seriously though, @SgtMate maybe you could provide some pseudo code or a draft PR to show where the new command fits in and what input/output would be involved in conditions.

harrisonablack:
> another thing i need to add to my request is some kind of dry-run functionality. where only the resulting download candidate is returned without downloading it. I need this to check if an update is even required once the container is restarted. Currently im just comparing the old and new filename. This would greatly help in reducing unneeded downloads.
> 
> @harrisonablack you think you can implement this with githubs given api?

@SgtMate As well, i'm assuming you want some *machine readable* format for the output? Or is human readable fine?

I would look at using [`ResultsFileWriter`](https://github.com/itzg/mc-image-helper/blob/main/src/main/java/me/itzg/helpers/files/ResultsFileWriter.java). The tool gives docker-minecraft-server a list of Enviroment variables to set. We could then create a command, `mc-image-helper install-gtnh --check-updates` which would return a ResultsFile containing a env var stating the output; `GTNH_UPDATES=TRUE` or something similar

harrisonablack:
@itzg Added two arguments, `--print-artifact` and `--no-download`. Currently `--print-artifact` logs artifact name using a `sl4j` log

```java
if (printArtifact) {
    log.info("Resolved artifact {}", candidate.block().getName());
}

```

For example, running

```
./gradlew run --args="github download-artifact --workflow daily-modpack-build.yml --artifact-pattern='GTNH-daily-\d{4}-\d{2}-\d{2}\+\d+-server-java17-25\.zip' --print-artifact --no-download GTNewHorizons/DreamAssemblerXXL"
```

returns

```
[mc-image-helper] 02:58:36.311 INFO  : Resolved artifact GTNH-daily-2026-07-17+628-server-java17-25.zip
```

itzg:
I would like it to be consistent with https://github.com/itzg/mc-image-helper#get where it's `--skip-existing` and `--output-filename`.

For `--output-filename` it needs to be a `System.out.println` so scripts can consume the output of the call like:

```
file=$(mc-image-helper github download-artifact --output-filename ...)
```

itzg:
...unless @SgtMate you had a different intention with the skipping of download or not.

harrisonablack:
@itzg

Fixed the print command.

My only concern with --skip-existing is how will it be used and when? 

For the GTNH instance, we will always modify the downloaded artifact to use it, unzip it for example. Then its not an easy "does this file already exist?", it comes back to the same thought of "Do we need a manifest"?

Otherwise if the thought of `--skip-existing` was only, "does this exact file already exist in the output dir?", thats easily implemented and am happy to do so!

itzg:
That'll be for @SgtMate to answer 😄

SgtMate:
**oh i see I`ve cause some discussions. Ill try to adress things one by one.**

> Seriously though, @SgtMate maybe you could provide some pseudo code or a draft PR to show where the new command fits in and what input/output would be involved in conditions.

I simply do a comparison of filenames from the downloaded zip file, if the name matches with the last version installed (stored in a file) the installation is skipped, if it does not match the update is triggered and if the file is missing a fresh install is triggered.

```
if [[ ! -f /data/.gtnh-version || "$(basename "$gtnh_download_path")" != "$(cat /data/.gtnh-version)" ]]; then
```
Here gtnh_download_path normaly stores the download location for the download and installation step. After download and install the value $gtnh_download is stored in the file which is derived from the download.
```
    if ! gtnh_download=$(mc-image-helper github download-artifact --output-filename -o=$base_dir --unzip $gtnh_download_options GTNewHorizons/DreamAssemblerXXL); then
      logError "Failed to download Daily Build for GTNH server files."
      exit 1
    fi
```

---
> I would lean more towards implementing a install-gtnh command in mc-image-helper to automate this process.

Yes I also see this as the best solution in the longway, but as this results in more complex migration from bash scripts over to java I'd like to split this. also because i think the now requested feature will be helpfull for other contributors of the minecraft server project, I try to have this be as usefull to the overall project as possible even tho I'm focussing on the GTNH implementation.

---

> ...unless @SgtMate you had a different intention with the skipping of download or not.

Close, --output-filename is important for me. The thing I need in addition is a way to just get the filename of the zip I would get back to compare to the last installed version before download.

---
> I would like it to be consistent with https://github.com/itzg/mc-image-helper#get where it's --skip-existing and --output-filename.

Agree, this also makes things easier to mainain from the script side same functionality same name.

---
> ...unless @SgtMate you had a different intention with the skipping of download or not.

No I dont use skip existing.

**I hope i was able to clarify all questions! Thanks for all the work you are putting in! ♥️**

harrisonablack:
@itzg 

Given that @SgtMate doesn't require `--skip-existing`, and his needed `--output-filename` is implemented, is there anything else you want to see changed or added before merge?

itzg:
I'm happy if @SgtMate is happy 😄 

I'll proceed with code review from here.

SgtMate:
I think Im happy, if something is still missing Ill adress this in a separate issue. Lets just get things implemented for now and improve incrementally.

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        try (SharedFetch sharedFetch = Fetch.sharedFetch("github download-artifact", sharedFetchArgs.options())) {
+            final GithubClient client = new GithubClient(sharedFetch, parent.apiBaseUrl, parent.token);
+
+            Mono<Artifact> candidate = resolveArtifact(client)
+                .switchIfEmpty(Mono.error(new GenericException("Github client failed to find an artifact")));
+
+            if (outputFilename) {
+                System.out.println(candidate.block().getName());
```
itzg: With reactive programming doing a `block` to grab a value is usually a code smell. Down on L123 you do another `block`. Since `block` initiates a subscription, it's basically saying "execute the reactive pipeline". So if someone picks `outputFilename` then you're executing the pipeline multiple times, which includes the `resolveArtifact` operation.

Instead, I'd say attach a `doOnNext` and put the conditional and println in there.

Before the `return` on L109 you can do a `block` since that's a single termination point.

Otherwise the block will get executed along with that `block` on L123.

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
```diff
@@ -0,0 +1,211 @@
...
+            if (noDownload) {
+                return ExitCode.OK;
+            }
+
+            // Github API requires a token with "'Actions' repository
+            // permissions (read)" to download an artifact
+            // https://docs.github.com/en/rest/actions/artifacts?apiVersion=2026-03-10#download-an-artifact
+            if (parent.token == null) {
+                throw new IllegalArgumentException("Must provide a github token to query artifact data");
+            }
```
itzg: Once you're in reactive, stay reactive. This could be done in an attached `flatMap` where the error case would return a `Mono.error(...)`

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
```diff
@@ -0,0 +1,211 @@
...
+
+            Path download = candidate
+                .doOnNext(artifact -> {
+                    log.info("Downloading artifact {}", artifact.getName());
+                }) 
+                .flatMap(artifact -> downloadArtifact(client, artifact)).block();
```
itzg: Since you might be doubting yourself now, this `block` is good. Putting on its own newline would be good for clarity though.

```suggestion
                .flatMap(artifact -> downloadArtifact(client, artifact))
                .block();
```

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
```diff
@@ -0,0 +1,211 @@
...
+
+
+            if (unzip) {
+                log.info("Unzipping artifact");
+                unzipArtifact(download, outputDirectory, overwrite);
+                Files.deleteIfExists(download);
```
itzg: When would it not exist?

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
```diff
@@ -0,0 +1,211 @@
...
+
+            if (noDownload) {
+                return ExitCode.OK;
+            }
+
+            // Github API requires a token with "'Actions' repository
+            // permissions (read)" to download an artifact
+            // https://docs.github.com/en/rest/actions/artifacts?apiVersion=2026-03-10#download-an-artifact
+            if (parent.token == null) {
+                throw new IllegalArgumentException("Must provide a github token to query artifact data");
```
itzg: These `IllegalArgumentException`'s throughout should be `InvalidParameterException`

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
```diff
@@ -0,0 +1,211 @@
...
+        try {
+            Files.createDirectories(extractionRoot);
+
+            try (ZipInputStream zis = new ZipInputStream(Files.newInputStream(zipPath))) {
+                ZipEntry entry;
+
+                while ((entry = zis.getNextEntry()) != null) {
+                    final Path output = extractionRoot.resolve(entry.getName()).normalize();
+
+                    if (!output.startsWith(extractionRoot)) {
```
itzg, line 186: When would this ever be false since the path you is resolved from `extractionRoot` on L182

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
```diff
@@ -0,0 +1,211 @@
...
+                        } else if (Files.notExists(output)) {
+                            Files.copy(zis, output);
+                        }
+                    }
+
+                    zis.closeEntry();
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to extract artifact archive", e);
```
itzg: Use a `GenericException` as a more specific version of `RuntimeException`

src/main/java/me/itzg/helpers/github/GithubClient.java
```diff
@@ -86,4 +94,113 @@ public Mono<Path> downloadLatestAsset(String org, String repo, @Nullable Pattern
...
+                        "/repos/{org}/{repo}/actions/workflows/{workflow}/runs",
+                        QueryParameters.queryParameters()
+                                .add("status", "success")
+                                .add("per_page", "1"),
+                        org, repo, workflow))
+                .withAuthorization("Bearer", token)
+                .toObject(WorkflowRunsResponse.class)
+                .assemble()
+                .flatMapIterable(WorkflowRunsResponse::getWorkflowRuns)
+                .next()
```
itzg: A very subtle reactive thing: no need to spin out a flux here when there's no I/O happening. Just do a "regular" operation:


```suggestion
                .mapNotNull(resp -> resp.getWorkflowRuns().stream().findFirst().orElse(null))
```

itzg (changes requested):
Just some minor things.

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
harrisonablack: This is known as a [ZipSlip](https://developer.android.com/privacy-and-security/risks/zip-path-traversal) attack. Where due to zip encoding, a file can contain path traversals, e.g. "../../../file.sh", then when normalised, it may escape the expected output folder.


> *"To mitigate this issue, before extracting each entry, you should always verify that the target path is a child of the destination directory."*

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
harrisonablack: Yeah, good point

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
harrisonablack: Yeahh, Reactor and reactive programming will take a minute to get used to. Reminds me of typescript i've had to write for classes; `Promise<T>`. It'll take a bit to fully get used to.

src/main/java/me/itzg/helpers/github/DownloadArtifactCommand.java
itzg: Ah, I missed the `normalize` at the end of the other. That's why I like putting chained method calls on a line each. 

Good to know about the attack vector. Later I'll refactor this out so other places can use it safely.

itzg (approved):
Thanks!

itzg:
Released in https://github.com/itzg/mc-image-helper/releases/tag/1.63.0

harrisonablack:
> Thanks!

All good! Anything else you want me to work on, just mention me in the issue or PR and I'll be there!

itzg:
Pulled into the image https://github.com/itzg/docker-minecraft-server/pull/4184 @SgtMate

PR #823, opened 2026-07-30
cf: Add regex option for Curseforge filename matching

harrisonablack:
refs: itzg/docker-minecraft-server#3686

### DIFF
--- a/src/main/java/me/itzg/helpers/curseforge/model/CurseForgeFile.java
+++ b/src/main/java/me/itzg/helpers/curseforge/model/CurseForgeFile.java
@@ -1,6 +1,7 @@
 package me.itzg.helpers.curseforge.model;
 
 import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import java.time.Instant;
 import java.util.List;
 import lombok.Data;
@@ -9,6 +10,7 @@ import lombok.Data;
 @JsonTypeName("File")
 public class CurseForgeFile {
 	private int gameId;
+	@JsonProperty("isAvailable")
 	private boolean isAvailable;
 	private String fileName;
 	private List<String> gameVersions;
@@ -21,6 +23,7 @@ public class CurseForgeFile {
 	private List<FileModule> modules;
 	private List<FileDependency> dependencies;
 	private long fileFingerprint;
+	@JsonProperty("isServerPack")
 	private boolean isServerPack;
 	private FileReleaseType releaseType;
 	private List<FileHash> hashes;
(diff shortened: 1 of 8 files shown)


itzg:
Like the issue description, I would like it to look for slash at start and end to indicate regex mode. With that users can indicate full match with the $ and ^ operators.

itzg:
Proposal: when the value is a slash-surrounded string then it will be processed as a regex such as

```
/.*1\.5$/
```

src/main/java/me/itzg/helpers/curseforge/model/CurseForgeFile.java
```diff
@@ -21,6 +23,7 @@ public class CurseForgeFile {
 	private List<FileModule> modules;
 	private List<FileDependency> dependencies;
 	private long fileFingerprint;
+	@JsonProperty("isServerPack")
```
harrisonablack, line 26: Both booleans would previously always be false, as they where not mapped properly by Jackson, Jackson would strip "is" prefix when mapping.

During `CurseForgeApiClient.GetModFileResponse`, we previously intended to filter out all "serverPacks" (as seen in comment in function), however due to the flag never mapping properly, we wouldn't actually skip "serverPacks"

Added test case to test skipping "serverPacks"

itzg: Thanks. I guess it has worked out fine since it later flags if the modpack zip in general is missing the needed metadata.

itzg:
Ping me when this is ready for review since GitHub doesn't seem to notify when leaving draft state.

harrisonablack:
> Ping me when this is ready for review since GitHub doesn't seem to notify when leaving draft state.

Should be good right now @itzg!

Just updating documentation

itzg:
Ping me when docs are updated.

harrisonablack:
> Ping me when docs are updated.

@itzg Just updated README and command description

itzg:
Thanks for the review reminder 😄

itzg:
The usage of `Optional` felt a bit awkward, which is actually a common "flaw" of `Optional`. Rather than try to explain what I was thinking via comments, I went and pushed my ideal encapsulation. It needs a few unit tests, but wanted to have you pondering that for now.

harrisonablack:
@itzg I added back the `try catch` on the Pattern.compile (assume removing that wasn't intentional), and then a few unit tests for `MultiMatcher`, aswell as an integration test between `CurseForgeApiClient` and `MultiMatcher`

itzg:
> I added back the try catch on the Pattern.compile (assume removing that wasn't intentional), 

Thanks! Indeed it wasn't intentional. 

> and then a few unit tests for MultiMatcher, aswell as an integration test between CurseForgeApiClient and MultiMatcher

Fantastic.

itzg (approved):
Great teamwork on this one 😄

Issue #824, opened 2026-07-30
patch: unreadable *_FILE placeholder source silently discards the whole patch


OowhitecatoO:
### Summary

When a `CFG_*_FILE` placeholder source cannot be read (for example, the path is a directory), `patch`:

1. discards the entire patch for that target — including operations that had already been applied — while still exiting `0`, and
2. reports the **target** file as unreadable, instead of the file it actually failed to read.

Verified against released `1.63.1`; the same code path is present on `main` (`3733cbb`), i.e. after #722 / #723.

### Reproduction

```bash
mkdir -p /tmp/repro/defs /tmp/repro/secret-as-dir
cd /tmp/repro

cat > target.yml <<'EOF'
proxies:
  velocity:
    enabled: false
    secret: ""
EOF

cat > defs/test.json <<'EOF'
{
  "file": "/tmp/repro/target.yml",
  "ops": [
    { "$set": { "path": "$.proxies.velocity.enabled", "value": "true", "value-type": "bool" } },
    { "$set": { "path": "$.proxies.velocity.secret",  "value": "${CFG_MY_SECRET}" } }
  ]
}
EOF

# CFG_MY_SECRET_FILE points at a directory — see "How this happens in practice"
CFG_MY_SECRET_FILE=/tmp/repro/secret-as-dir mc-image-helper patch /tmp/repro/defs
echo "rc=$?"
cat target.yml
```

### Observed

```
WARN : Failed to read content of /tmp/repro/target.yml: Is a directory
rc=0
```

```yaml
proxies:
  velocity:
    enabled: false     # op 1 was applied in memory, then thrown away
    secret: ""
```

`/tmp/repro/target.yml` is a readable regular file. The directory is `/tmp/repro/secret-as-dir`,
which the message never mentions.

### Compare with the missing-variable path

The same patch, with `CFG_MY_SECRET_FILE` and `CFG_MY_SECRET` simply unset:

```
WARN : Unable to set '$.proxies.velocity.secret' due to missing environment variables: [CFG_MY_SECRET]
rc=0
```

```yaml
    enabled: true      # op 1 kept
```

| `CFG_MY_SECRET` state | message names | rc | op 1 written? |
|---|---|---|---|
| unset | the operation and the variable | 0 | yes |
| `_FILE` set, source unreadable | the **target** file | 0 | **no** |

### Root cause

`Interpolator.readValueFromFile` (`Interpolator.java:77-80`) lets the `IOException` from
[...]

PR #825, opened 2026-07-30
patch: name the file that could not be read for a placeholder value

OowhitecatoO:
Addresses the reporting half of #824. Semantics are deliberately left unchanged, so this can go
in independently of the behaviour change proposed there.

### Problem

`Interpolator.readValueFromFile` let the `IOException` from `Files.readAllBytes` propagate bare.
It is not a `MissingVariablesException`, so it was not handled at `PatchSetProcessor.java:186` and
instead surfaced at the `catch (IOException)` around the whole read-apply-write block. That
handler assumes the failure concerned the patch target, so the warning named the target file —
which is usually perfectly readable — and never mentioned the file that actually failed.

With `CFG_MY_SECRET_FILE` pointing at a directory (see the issue for why this is a common
accident), released 1.63.1 reports:

```
WARN : Failed to read content of /tmp/repro/target.yml: Is a directory
```

`target.yml` is a readable regular file. Nothing in the log points at the placeholder source.

### Change

- Wrap the failure in `readValueFromFile` with the file name and the underlying reason.
- Reword the warning in `processPatch` so it no longer claims the target was unreadable — the
  same `catch` also covers I/O performed while applying the ops.

The same run now reports:

```
WARN : Failed to apply patch from /tmp/repro/defs/test.json to /tmp/repro/target.yml: \
       Failed to read placeholder value from /tmp/repro/secret-as-dir (Is a directory)
```

The patch definition, the target and the file that actually failed are all named.

### Not changed

The patch is still skipped with a warning and exit code 0, and the ops that were applied before
the failure are still discarded. #824 proposes making that case fail outright instead; this PR
deliberately does not change it.

### Verified

- `./gradlew test` — 488 tests across 74 classes, all green (JDK 21 toolchain).
- The added test `InterpolatorTest.reportsFileNameWhenValueFileCannotBeRead` fails on `main`
  without the `Interpolator` change and passes with it (checked both ways).
- End-to-end via `installDist`, reproducing with a `*_FILE` variable pointing at a directory.

### DIFF
--- a/src/main/java/me/itzg/helpers/env/Interpolator.java
+++ b/src/main/java/me/itzg/helpers/env/Interpolator.java
@@ -75,8 +75,17 @@ public class Interpolator {
     }
 
     private String readValueFromFile(String filename) throws IOException {
-        final String content = new String(Files.readAllBytes(Paths.get(filename)), StandardCharsets.UTF_8);
-        return content.trim();
+        final byte[] content;
+        try {
+            content = Files.readAllBytes(Paths.get(filename));
+        } catch (IOException e) {
+            // Without the file name here, the failure gets reported against whatever file the
+            // caller was processing, which is not the file that could not be read.
+            throw new IOException(
+                String.format("Failed to read placeholder value from %s (%s)", filename, e.getMessage()), e
+            );
+        }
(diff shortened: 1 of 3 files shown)


itzg:
Went with #826

PR #828, opened 2026-07-31
Accept standalone patch definition files

### LINKED ISSUE #827: Patch operation should auto detect patch vs patch set files
That would have addressed things like the issue resolved in https://github.com/itzg/docker-minecraft-server/pull/4201

Guflly:
Detect whether a JSON file contains a PatchSet or a single PatchDefinition. Existing PatchSet and directory handling stays unchanged.

Tests: `./gradlew check --no-problems-report`

Closes #827

### DIFF
--- a/README.md
+++ b/README.md
@@ -1163,8 +1163,8 @@ Supports the file formats:
 - JSON5
 - Yaml
 - TOML, but processed output is not pretty
-      FILE_OR_DIR   Path to a PatchSet json file or directory containing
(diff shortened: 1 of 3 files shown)


itzg (approved):
Thanks for jumping on this so quickly.

itzg:
Releasing in https://github.com/itzg/mc-image-helper/releases/tag/1.64.0 and then dependency bot will apply to the images.

PR #831, opened 2026-08-01
cf: load CurseForge API key from file env

### LINKED ISSUE #829: Read CF_API_KEY_FILE directly rather than excepting wrapper script to load into CF_API_KEY
Following up to https://github.com/itzg/docker-minecraft-server/pull/4208 the loading of `CF_API_KEY` env var from `CF_API_KEY_FILE` really should be done within mc-image-helper. With that these locations can be eliminated:

- https://github.com/itzg/docker-minecraft-server/blob/d85904036f13b4bc2b65b02baf438f9061aa1363/scripts/start-deployAutoCF#L33
- https://github.com/itzg/docker-minecraft-server/blob/d85904036f13b4bc2b65b02baf438f9061aa1363/scripts/start-setupModpack#L21

tomatotomata:
## Summary

Load the CurseForge API key directly from the conventional CF_API_KEY_FILE environment variable.

## Details

- Added CF_API_KEY_FILE alongside the existing CF_API_KEY constant
- Preserve explicit --api-key precedence
- Read the file as UTF-8 and trim the trailing newline commonly present in mounted secrets
- Treat missing, unreadable, or empty files as configuration errors without logging the secret
- Added focused tests for file loading, precedence, and missing-file handling

## Validation

- ./gradlew test --tests 'me.itzg.helpers.curseforge.*'`n- git diff --check`n
Closes #829

### DIFF
--- a/src/main/java/me/itzg/helpers/curseforge/ApiKeyHelper.java
+++ b/src/main/java/me/itzg/helpers/curseforge/ApiKeyHelper.java
@@ -1,5 +1,9 @@
 package me.itzg.helpers.curseforge;
 
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Map;
 import lombok.extern.slf4j.Slf4j;
 import me.itzg.helpers.errors.InvalidParameterException;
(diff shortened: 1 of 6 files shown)


src/main/java/me/itzg/helpers/curseforge/ApiKeyHelper.java
```diff
@@ -46,11 +50,34 @@ EXPECTED_API_KEY_PREFIX, partiallyRedactApiKey(apiKey)
      * @throws InvalidParameterException if no API key is provided or cannot be loaded
      */
     public static String loadApiKey(String providedApiKey) {
+        return loadApiKey(providedApiKey, System.getenv());
```
itzg: Accessing the system env is a code smell. I would be interested in a new command line arg for the api key file and localize the env loading there like the api key argument.

itzg (changes requested):
Thank you for quickly addressing this. I'm hoping you'll find the requested change is doable but let me know if that turns out messier than expected.

tomatotomata:
I localized the file loading behind a new `--api-key-file PATH` option on both CurseForge commands and pushed `8dc3a11`.

The option defaults from `CF_API_KEY_FILE`, preserves explicit `--api-key` precedence, and passes the resolved path into the installer/helper instead of having the helper reach into the process environment. The focused helper tests now cover the path-based loader and missing-file error.

`compileJava` completed successfully and `git diff --check` passes. The focused test task could not resolve WireMock metadata because this environment cannot reach Maven Central, so CI still needs to run the tests.

Let me know what you think.

itzg (approved):
Thanks!

PR #837, opened 2026-08-03
patch: only warn when the declared file format is not supported

OowhitecatoO:
`resolveFileFormat` logs "The file format X is not supported" unconditionally whenever a patch declares `file-format`, so a valid declaration warns even though the patch is applied correctly. The `else` branch already guards its warning on the lookup having failed; this makes the `if` branch match it.

A parameterized test comes with it, covering both directions: a supported name declared on a file whose suffix cannot be detected applies the patch and logs no warning, and an unsupported name leaves the file alone and still warns. The first case fails on current `main`.

Unrelated but adjacent: `patch` has no properties format, so a definition pointed at a `.properties` file is skipped with a warning and exit 0. I hit this on a mod's own config, which `set-properties` does not cover. A prototype using `java.util.Properties` in both directions works; happy to send it if you are interested.

<details><summary>Reproduction, if useful</summary>

```json
{
  "patches": [
    {
      "file": "t.yml",
      "file-format": "yaml",
      "ops": [ { "$set": { "path": "$.motd", "value": "patched" } } ]
    }
  ]
}
```

```
$ mc-image-helper patch patch.json
WARN : The file format yaml is not supported
$ cat t.yml
motd: "patched"
```

The patch went through; only the message is wrong.

</details>

### DIFF
--- a/src/main/java/me/itzg/helpers/patch/PatchSetProcessor.java
+++ b/src/main/java/me/itzg/helpers/patch/PatchSetProcessor.java
@@ -115,7 +115,9 @@ public class PatchSetProcessor {
         final FileFormat fileFormat;
         if (formatName != null) {
             fileFormat = getFileFormatByName(formatName);
-            log.warn("The file format {} is not supported", formatName);
+            if (fileFormat == null) {
+                log.warn("The file format {} is not supported", formatName);
+            }
         } else {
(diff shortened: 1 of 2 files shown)


itzg (approved):
Thanks for even more great cleanup

PR #842, opened 2026-08-03
patch: support properties files

OowhitecatoO:
`patch` has no properties format, so a definition pointed at a `.properties` file is skipped with a warning and exit 0. This is the change I mentioned on #837. The case I ran into is a mod's own config — Simple Voice Chat's `voicechat-server.properties` — which `set-properties` does not cover, since that one is driven by `server.properties` definitions.

The model is flat and both directions go through `java.util.Properties`, the same pair `SetPropertiesCommand` already relies on, so no escaping is reimplemented here. Comments are not retained, as with `set-properties`. A value a properties file cannot hold — an object, an array, a `list of ...`, a null — now fails the patch with exit 2 rather than being written out as `list=[1, 2]`. That applies only to properties files; the other formats are unchanged.

One change lands outside the new format: `PatchSetProcessor` now encodes before opening the file, since `Files.newOutputStream` truncates on open and an exception out of `encode` used to leave a zero-byte file behind. Properties is the first format whose encode can fail, so this is latent today rather than something anyone can hit, but it belongs with the feature that makes it reachable.

<details><summary>Details, if useful</summary>

**Why the model is flat.** Keys are taken exactly as they appear in the file, with no dot-splitting into nested objects. `server.properties` carries `query.port` and `rcon.port` alongside plain keys, and splitting on dots would collide if a file ever held both `rcon` and `rcon.port`. A key containing a dot therefore needs bracket notation, `$['query.port']`, and getting that wrong is a hard failure rather than a silent miss. Hyphens are fine in plain dot notation, so `$.max-players` works.

**Key order** is retained from the original file, which `Properties.store` on its own does not do. That is the only thing added on top of the JDK.

**No warning for the lost comments.** Documenting it seemed better than a message on every patched file, and in the case that prompted this the mod rewrites its own comments from code on the next start.

**Docs.** The `patch` usage text and the `PatchDefinition.file-format` description are updated; the latter listed only "json, yaml", so it was already missing json5 and toml. I edited the affected lines of the generated README block by hand instead of regenerating it, because the committed block is about 486 lines out of sync with what the tool emits today and regenerating would bury this change in unrelated churn.

**Verified.**

- `./gradlew test` — 549 tests, all green, JDK 21 toolchain.
- `PropertiesFileFormatTest` round-trips 24 awkward entries in both directions: `=`, `:`, `#`, `!`, backslashes, newlines, tabs, leading and trailing spaces, control characters, empty values, latin-1, multi-byte and supplementary-plane characters, and the same set again in keys. Each is also read back from what `Properties.store` itself writes, so escaping parity with `set-properties` is pinned rather than assumed.
- Both new behaviours were checked red first: `setInProperties` fails if the format is not registered, and `leavesFileAloneWhenItCannotBeEncoded` fails if the encode-before-open reorder is reverted.
- End to end against the real Simple Voice Chat config: three ops applied with `value-type` `int`, `float` and `bool`; a `list of string` op rejected with exit 2 and the file left untouched, its 33 comment lines included.

</details>

### DIFF
--- a/src/main/java/me/itzg/helpers/patch/PropertiesFileFormat.java
+++ b/src/main/java/me/itzg/helpers/patch/PropertiesFileFormat.java
@@ -79,8 +79,15 @@ public class PropertiesFileFormat implements FileFormat {
         final StringWriter out = new StringWriter();
         properties.store(out, null);
 
-        // store() leads with the date, which java.properties.date can turn into several lines
-        final String written = out.toString();
+        return stripLeadingComments(out.toString());
+    }
+
+    /**
+     * {@link Properties#store} leads with the date, which the java.properties.date system
+     * property can turn into several lines. Lines are located by their newline, so this works
+     * whichever line separator {@link Properties#store} used.
+     */
+    static String stripLeadingComments(String written) {
         int start = 0;
         while (start < written.length()
             && (written.charAt(start) == '#' || written.charAt(start) == '!')) {
--- a/src/main/java/me/itzg/helpers/patch/PropertiesFileFormat.java
+++ b/src/main/java/me/itzg/helpers/patch/PropertiesFileFormat.java
@@ -1,0 +1,112 @@
+package me.itzg.helpers.patch;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import java.io.IOException;
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.util.AbstractMap;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import me.itzg.helpers.errors.GenericException;
+import me.itzg.helpers.errors.InvalidParameterException;
+
+/**
+ * Properties files are a flat namespace of string values, so the decoded model is a flat map
+ * of the keys exactly as they appear in the file. All escaping is delegated to
+ * {@link Properties}; the only thing added here is preserving the key order of the original
+ * file, which {@link Properties#store} would otherwise lose.
+ * <p>
+ * Comments are not retained, which matches the {@code set-properties} command.
+ */
+public class PropertiesFileFormat implements FileFormat {
+
+    private static final String[] SUFFIXES = {"properties"};
+
+    @Override
+    public String[] getFileSuffixes() {
+        return SUFFIXES;
+    }
+
+    @Override
+    public String getName() {
+        return "properties";
+    }
+
+    @Override
+    public Map<String, Object> decode(String content) throws IOException {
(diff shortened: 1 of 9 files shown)


src/main/java/me/itzg/helpers/patch/PropertiesFileFormat.java
```diff
@@ -0,0 +1,112 @@
...
+
+        final StringWriter out = new StringWriter();
+        properties.store(out, null);
+
+        // store() leads with the date, which java.properties.date can turn into several lines
+        final String written = out.toString();
+        int start = 0;
+        while (start < written.length()
+            && (written.charAt(start) == '#' || written.charAt(start) == '!')) {
+            final int endOfLine = written.indexOf('\n', start);
```
itzg, line 94: Is this portable with Windows?

src/test/java/me/itzg/helpers/patch/PropertiesFileFormatTest.java
```diff
@@ -0,0 +1,123 @@
...
+    @Test
+    void retainsKeyOrder() throws IOException {
+        final Map<String, Object> original = new LinkedHashMap<>();
+        original.put("zz", "1");
+        original.put("aa", "2");
+        original.put("mm", "3");
+
+        // store() separates entries with the platform line separator
+        final String eol = System.lineSeparator();
+        assertThat(format.encode(original)).isEqualTo("zz=1" + eol + "aa=2" + eol + "mm=3" + eol);
```
itzg: FYI I believe AssertJ has an operator for comparing ignoring line endings.

itzg (commented):
Thanks. Just a question to answer.

src/main/java/me/itzg/helpers/patch/PropertiesFileFormat.java
OowhitecatoO: It is. `indexOf('\n')` finds the terminator whether the line ends in LF or CRLF, and the `\r` goes with it since the substring starts after the newline — so no stray carriage return is left behind. Rather than leave that as an assertion I pulled the loop out into `stripLeadingComments` and fed it both forms directly, including a multi-line date and an unterminated one.

The entries themselves do keep whatever separator `Properties.store` used, so on Windows a patched file comes out CRLF. That is the same as the json and json5 formats already behave, since Jackson's default indenter reads `line.separator`; yaml and toml always emit LF. I left properties matching the json side rather than normalising, but say the word if you would rather it always wrote LF.

src/test/java/me/itzg/helpers/patch/PropertiesFileFormatTest.java
OowhitecatoO: `isEqualToNormalizingNewlines`, thank you — that reads much better than concatenating `System.lineSeparator()`.

itzg (approved):

PR #847, opened 2026-08-12
mcopy: 403 error handling

### LINKED ISSUE #319: mcopy: simplify 403 failures
Such as

[message.txt](https://github.com/itzg/mc-image-helper/files/12811587/message.txt)

[From Discord](https://discord.com/channels/660567679458869252/660569641550217327/1159311385898000416)

harrisonablack:
fixes #319

#649 improved the primary `FailedRequestException` output by simplifying the logged error instead of full stack trace, however it didn't address the `onErrorDropped` issue

What happened in #319, when `MultiCopyCommand` is called, it takes a list of sources, in this case CurseForge URLs, and converts them into a Flux Stream.

A `flatMap` is then performed on that Flux Stream, which subscribes to multiple sources concurrently.

When one download returned a 403, the merged Flux propogated that error and cancelled the other running downloads. If another download during that time reported its own 403, Reactor couldn't deliver it to the subscriber, as they have already terminated, and instead it was routed to `onErrorDropped`, this is why in the issue we see the stack trace containing the response body, headers, etc logged. 

`flatMap` has been replaced with `concatMap` for source processing. `concatMap` subscribes to one source at a time, waits for it to complete before subscribing to the next. Errors are propagated immediately after each completion and later sources not started.

We could also use a `flatMapDelayError`, which would preserve parallel downloads while delaying error propagation until active sources complete, however i lean towards the fail-fast of `concatMap`

A test has been added which asserts that a second remote source is not requested after the first source fails.

### DIFF
--- a/src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
+++ b/src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
@@ -41,7 +41,7 @@ import reactor.core.scheduler.Schedulers;
 @Command(name = "mcopy", description = "Multi-source file copy operation with with managed cleanup. "
     + "Supports auto-detected sourcing from file list, directories, and URLs")
 @Slf4j
-public class MulitCopyCommand implements Callable<Integer> {
+public class MultiCopyCommand implements Callable<Integer> {
     @SuppressWarnings("unused")
     @Option(names = {"--help", "-h"}, usageHelp = true)
     boolean showHelp;
--- a/src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
+++ b/src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
@@ -100,7 +100,7 @@ public class MultiCopyCommand implements Callable<Integer> {
             final List<Path> results = Flux.fromIterable(sources)
                 .map(String::trim)
                 .filter(s -> !s.isEmpty())
-                .flatMap(source -> processSource(sharedFetch, source, fileIsListingOption, dest))
+                .concatMap(source -> processSource(sharedFetch, source, fileIsListingOption, dest))
                 .collectList()
                 .block();
 
@@ -185,7 +185,7 @@ public class MultiCopyCommand implements Callable<Integer> {
                     final List<String> lines = Files.readAllLines(path);
                     return Flux.fromIterable(lines)
                         .filter(this::isListingLine)
-                        .flatMap(src -> processSource(sharedFetch, src,
+                        .concatMap(src -> processSource(sharedFetch, src,
                             // avoid recursive file-listing processing
                             false,
                             destination
@@ -326,7 +326,7 @@ public class MultiCopyCommand implements Callable<Integer> {
                     .flatMapMany(content -> Flux.just(content.split("\\r?\\n")))
                     .filter(this::isListingLine)
             )
-            .flatMap(url -> processSource(sharedFetch, url, false, destination))
+            .concatMap(url -> processSource(sharedFetch, url, false, destination))
             .doOnTerminate(sharedFetch::close)
             .checkpoint("Processing remote listing at " + source, true);
     }
--- a/src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
+++ b/src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
@@ -108,18 +108,15 @@ public class MultiCopyCommand implements Callable<Integer> {
 
         try (SharedFetch sharedFetch = Fetch.sharedFetch("mcopy", sharedFetchArgs.options())) {
 
-
             final List<Path> results = Flux.fromIterable(sources)
                 .map(String::trim)
                 .filter(s -> !s.isEmpty())
                 .flatMapDelayError(
-                    source -> Flux.defer(() -> processSource(sharedFetch, source, fileIsListingOption, dest))
-                        .doOnError(error ->
-                            log.error("Failed to process source {}: {}", source, error.getMessage())
-                        ),
+                    source -> processSource(sharedFetch, source, fileIsListingOption, dest),
                     maxConccurentSources,
                     PREFETCH
                 )
+                .doOnError(error -> log.error("Failed to process source: {}", error.getMessage()))
                 .collectList()
                 .block();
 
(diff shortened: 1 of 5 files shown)


harrisonablack:
Aswell, "MultiCopyCommand.java" and "MultiCopyCommandTest.java" was misspelled as "MulitCopyCommand.java" and "MulitCopyCommandTest.java" this has been fixed

itzg:
Thanks! The explanation of the issue makes total sense now. 

I would be interested in you trying flatMapDelayError. I was explicitly wanting the downloads to occur concurrently. It was one of the main reasons I chose Netty Reactive since expressing that as streams works out nicely.

itzg:
...actually the original flatMap isn't fully concurrent since I later learned the variant with concurrency amount is needed

https://github.com/itzg/mc-image-helper/blob/9efe522224be4d7c76bf63bfedcd12ed2b83366d/src/main/java/me/itzg/helpers/curseforge/CurseForgeInstaller.java#L589

Looks like there's also a method variant that accepts `delayError`. 

Overall, I do agree about the fail-fast spirit. It would be nice if it could go parallel, properly handle the first error, and not trip on itself for any others.

harrisonablack:
> I would be interested in you trying flatMapDelayError. I was explicitly wanting the downloads to occur concurrently. It was one of the main reasons I chose Netty Reactive since expressing that as streams works out nicely.

@itzg 

I've replaced `concatMap` with `flatMapDelayError`, now maintains concurrent downloads. 

I've also added a Reactor `CompositeException` handler so propagated CompositeExceptions follow the format of existing Exception logs in `ExceptionHandler.java`.

One other note, `flatMapDelayError` requires concurrency and prefetch defined, are we happy hard coding them at sane defaults? Or want to expose them as Picocli args?

itzg:
> are we happy hard coding them at sane defaults? Or want to expose them as Picocli args?

Hardcoding would be fine, but since we're already right there in a command class it'll be super easy to declare args and the defaults there on the field initializer. 

**Edit**

The prefetch doesn't need to be an arg since that's too low level. Declare a final static (constant) for that though to avoid magic values.

itzg:
> I've also added a Reactor CompositeException handler so propagated CompositeExceptions follow the format of existing Exception logs in ExceptionHandler.java.

Fantastic! I didn't know that was a thing.

harrisonablack:
> The prefetch doesn't need to be an arg since that's too low level. Declare a final static (constant) for that though to avoid magic values.

Ahh i just saw that after i commited it :/

Reversed it and added a final static int for prefetch

harrisonablack:
@itzg Let me know if you need anything else changed around!

src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
```diff
@@ -93,14 +93,33 @@ public void setSources(List<String> sources) {
...
+    private final static int PREFETCH = 1;
+
+    @Option(names = "--max-concurrent-sources", defaultValue = "10", description = "Maximum number of sources to process concurrently")
+    int maxConccurentSources;
+
+
     @Override
     public Integer call() throws Exception {
 
+        if (maxConccurentSources <= 0) {
```
itzg, line 105: At some point I should add in Jakarta Validation and that'll integrate with picocli

src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
```diff
@@ -93,14 +93,33 @@ public void setSources(List<String> sources) {
...
+
         try (SharedFetch sharedFetch = Fetch.sharedFetch("mcopy", sharedFetchArgs.options())) {
+
+
             final List<Path> results = Flux.fromIterable(sources)
                 .map(String::trim)
                 .filter(s -> !s.isEmpty())
-                .flatMap(source -> processSource(sharedFetch, source, fileIsListingOption, dest))
+                .flatMapDelayError(
+                    source -> Flux.defer(() -> processSource(sharedFetch, source, fileIsListingOption, dest))
```
itzg: I'm not sure if the defer is needed here. Also could shift the on-error after the flat-map stage I believe.

src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
```diff
@@ -185,11 +204,17 @@ private Flux<Path> processListingFile(SharedFetch sharedFetch, Path listingFile,
                     final List<String> lines = Files.readAllLines(path);
                     return Flux.fromIterable(lines)
                         .filter(this::isListingLine)
-                        .flatMap(src -> processSource(sharedFetch, src,
-                            // avoid recursive file-listing processing
-                            false,
-                            destination
-                        ));
+                        .flatMapDelayError(
+                            src -> Flux.defer(() -> processSource(sharedFetch, src,
```
itzg: Just noting same here

itzg (commented):
Just pondering the `defer`

src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
harrisonablack: @itzg 

The Defer and `doOnError` was inside the `flatMapDelayError`, as when an error is propagated, the specific source is still available and can be logged. e.g

```java
.doOnError(error -> log.error("Failed to process source {}: {}", source, error.getMessage()))
```

```
[mc-image-helper] 23:01:14.591 ERROR : Failed to process source http://localhost:51794/first.jar: HTTP request of http://localhost:51794/first.jar failed with 403 Forbidden: Getting file
[mc-image-helper] 23:01:14.591 ERROR : Failed to process source http://localhost:51794/second.jar: HTTP request of http://localhost:51794/second.jar failed with 403 Forbidden: Getting file
[mc-image-helper] 23:01:14.592 ERROR : 'mcopy' command failed. Version is null: FailedRequestException: HTTP request of http://localhost:51794/first.jar failed with 403 Forbidden: Getting file; FailedRequestException: HTTP request of http://localhost:51794/second.jar failed with 403 Forbidden: Getting file; Exception: #block terminated with an error
```

If we move the doOnError to outside the flatMapDelayError, only the error is propagated, and handled once, not per source.

```java
            final List<Path> results = Flux.fromIterable(sources)
                .map(String::trim)
                .filter(s -> !s.isEmpty())
                .flatMapDelayError(
                        source -> processSource(sharedFetch, source, fileIsListingOption, dest),
                        maxConccurentSources,
                        PREFETCH
                        )
                .doOnError(error -> log.error("Failed to process source: {}", error.getMessage()))
                .collectList()
                .block();
```

```
[mc-image-helper] 12:56:01.450 ERROR : Failed to process source: Multiple exceptions
[mc-image-helper] 12:56:01.450 ERROR : 'mcopy' command failed. Version is null: FailedRequestException: HTTP request of http://localhost:55078/first.jar failed with 403 Forbidden: Getting file; FailedRequestException: HTTP request of http://localhost:55078/second.jar failed with 403 Forbidden: Getting file; Exception: #block terminated with an error
```

However, i would agree with shifting the on-error and removing the defer, as the current logging prints each exceptions twice

src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
harrisonablack: > At some point I should add in Jakarta Validation and that'll integrate with picocli

I would very much agree

src/main/java/me/itzg/helpers/sync/MultiCopyCommand.java
harrisonablack: > I'm not sure if the defer is needed here. Also could shift the on-error after the flat-map stage I believe.

Pushed commit removing defer and moving `doOnError` after `flatMapDelayError`

itzg (approved):
Thanks again

PR #861, opened 2026-09-07
Add check-zip-slip and unzip helper commands

harrisonablack:
Adds two commands under `me.itzg.helpers.files.ZipCommand`, which provide a wrapper over `me.itzg.helpers.files.Zip#containsZipSlip` and `me.itzg.helpers.files.Zip#unzip`

```bash
./gradlew run --args="zip --help"
```
```text
Usage: mc-image-helper zip [-h] [COMMAND]
Zip related helpers, asserting on zip-slips and extracting zips
  -h, --help
Commands:
  check-zip-slip  Checks if a zip contains a zip-slip
  unzip           Unzip archive, checking for zip-slips
```

refs: #853

PR #863, opened 2026-09-09
Expand zip/zip-slip helper command

harrisonablack:
Expand upon #861 to replace bash Zip Slip checks added in itzg/docker-minecraft-server#4247

Moved `ZipCommand` to `ArchiveCommand`, renamed "check-zip-slip" and "unzip" to "check-path-traversal" and "extract".

Expanded supported types, now supports types from [start-utils](https://github.com/lbellows/docker-minecraft-server/blob/f0bdc9732ed34a7cc29c096b0febb0c9d068e99a/scripts/start-utils#L580-L601)
- application/zip
- application/x-tar
- application/gzip
- application/x-bzip2
- application/zstd

Added an abstract `TarArchive` class with a protected `InputStream`. This input stream is then overwritten in each subclass depending on the archive type.


src/main/java/me/itzg/helpers/files/archive/ArchiveCommand.java
```diff
@@ -0,0 +1,52 @@
...
+import picocli.CommandLine.ExitCode;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.Parameters;
+
+@Command(name = "archive", description = "Archive helpers, checking for path traversal and extracting archives")
+public class ArchiveCommand {
+    @Option(names = { "--help", "-h" }, usageHelp = true)
+    boolean help;
+
+    @Command(name = "check-path-traversal", description = "Checks if an archive contains path traversal")
```
itzg: Perhaps add the word "attack", borrowing from the intro of https://owasp.org/www-community/attacks/Path_Traversal

```suggestion
    @Command(name = "check-path-traversal", description = "Checks if an archive contains path traversal attack")
```

harrisonablack:
I initially liked the idea of a few simple static helpers, like the initial PR. But now, after realisation of having to bring in support for other types, it made more sense to me to construct support polymorphically.

Instead of just having a switch to several different static functions, i liked the idea of having an interface across the archive types.

src/main/java/me/itzg/helpers/files/archive/Archive.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+
+import org.apache.commons.compress.archivers.ArchiveException;
+
+public interface Archive {
```
itzg, line 8: With four default methods and a static, you've abused the meaning of an `interface`. Default interface method concept was added to the language mostly to allow for backward compatibility as interfaces added more methods. As such they tend to be wrappers that call upon the non-default methods. Instead these look like convenience methods for the implementations of the interface. 

I'm thinking just move those to an "ArchiveUtils" class leaving just the two interface methods.

src/main/java/me/itzg/helpers/files/archive/Archive.java
```diff
@@ -0,0 +1,66 @@
...
+        Files.createDirectories(output.getParent());
+
+        if (overwrite) {
+            Files.copy(input, output, StandardCopyOption.REPLACE_EXISTING);
+        } else if (Files.notExists(output)) {
+            Files.copy(input, output);
+        }
+    }
+
+    static Archive parseArchive(Path archive) throws IOException {
```
itzg: Move this one to something like `ArchiveFactory`. As it is you've inverted the abstraction where the interface needs to know of all the concrete types and be changed if others were added.

src/main/java/me/itzg/helpers/files/archive/TarBzip2.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class TarBzip2 extends TarArchive {
+
+    public TarBzip2(java.nio.file.Path archive) {
+        super(archive);
+    }
+
+    @Override
+    protected InputStream openInputStream() throws IOException {
+        return new BZip2CompressorInputStream(Files.newInputStream(archive()));
```
itzg, line 17: An idea to try out: instead of a subclass for each, perhaps pass TarArchive and "input stream opener" instance. It would be a functional interface that can be instantiated as a lambda by the new ArchiveFactory. 

It'll then be composable like the `tar` command itself is.

src/main/java/me/itzg/helpers/files/archive/TarArchive.java
```diff
@@ -0,0 +1,83 @@
...
+                    copyEntry(tar, output, overwrite);
+                }
+            }
+        }
+
+        return extractionRoot;
+    }
+
+    private String findUnsafeEntry() throws IOException {
+        final Path extractionRoot = archive.toAbsolutePath().normalize().getParent();
```
itzg, line 69: This extraction root seems a bit arbitrary, but is that because some kind of anchor reference point is needed and doesn't matter if it's not the intended output directory?

Perhaps add an inline comment to clarify.

src/main/java/me/itzg/helpers/files/archive/TarArchive.java
```diff
@@ -0,0 +1,83 @@
...
+            log.warn("Path traversal detected at: " + unsafeEntry + " in archive: " + archive.toAbsolutePath());
+            return true;
+        }
+
+        return false;
+    }
+
+    @Override
+    public Path extract(Path destination, boolean overwrite) throws IOException, ArchiveException {
+        final String unsafeEntry = findUnsafeEntry();
```
itzg, line 43: You're reading the archive twice with this. I know it's tempting to reuse the standalone entry method, but please find a way to do the evaluation inline for extraction.

src/main/java/me/itzg/helpers/files/archive/Archive.java
```diff
@@ -0,0 +1,66 @@
...
+        if (!Files.exists(archive)) {
+            throw new IllegalArgumentException("File does not exist: " + archive.toAbsolutePath());
+        } else if (!Files.isRegularFile(archive)) {
+            throw new IllegalArgumentException("File is not a regular file: " + archive.toAbsolutePath());
+        }
+
+        final String contentType = Files.probeContentType(archive);
+
+        if (contentType == null) {
+            throw new IOException("Failed to read file MIME type: " + archive.toAbsolutePath());
```
itzg: It might be worth adding a fallback to looks at the file extension. I seem to remember with the Alpine image variant the OS mime type detection is fairly sparse.

itzg (commented):
Thanks for this expansion (pun not originally intended).

itzg:
> I initially liked the idea of a few simple static helpers, like the initial PR. But now, after realisation of having to bring in support for other types, it made more sense to me to construct support polymorphically.
> 
> 
> 
> Instead of just having a switch to several different static functions, i liked the idea of having an interface across the archive types.

As you'll see from the comments there's a balancing act of polymorphism, composability, and acceptable use of utility methods. I appreciate you thinking out loud about some of the choices considered.

src/main/java/me/itzg/helpers/files/archive/Archive.java
harrisonablack: I guess my initial logic was to minimise boiler plate, if i could include the helpers in an already present package-wide way without creating another class.

However, your exactly right, that defeats the whole purpose of the interface to begin with.

Moved the functions to `ArchiveUtils`

src/main/java/me/itzg/helpers/files/archive/TarBzip2.java
harrisonablack: I actually really like this implementation. Definitely not something i would've approached the problem with architecturally before.

Whilst it does loose the inherit information of what the raw archive is, it doesn't really matter after the input stream is created.

src/main/java/me/itzg/helpers/files/archive/TarArchive.java
harrisonablack: > This extraction root seems a bit arbitrary, but is that because some kind of anchor reference point is needed and doesn't matter if it's not the intended output directory?

Pretty much exactly, it doesn't matter where the root is taken, because we aren't actually unzipping it, "*if* i unzipped this inplace, is it staying inside the parent?"

src/main/java/me/itzg/helpers/files/archive/TarArchive.java
harrisonablack: The initial thought is, "if i already unzipped half of the zip, then found a ZipSlip, what do i do?"

Ideally we wouldn't just leave the partially unzipped files

We could;
- Initially unzip to a TempDir, then detele
- Track the already extracted files, then delete

or

Initially pass over the zip, check for ZipSlips, then *IF* clear unzip. We wouldn't ever expect to be in a partially unzipped state due to a detected ZipSlip

I primarily landed on that decision due to the lower complexity, however as you note, the trade-off is iterating over the file twice.

If the priority is still to remove the second read, i'm happy to look at remove the initial `findUnsafeEntry`, and instead tracking unzipped files

itzg: Yeah, excellent point. I also just realized that traversing the entries is actually quite minimal vs a full extraction since it only needs to skip through metadata. 

Let's stick with what you have.

itzg (approved):
Thanks. Very thorough tests.

src/main/java/me/itzg/helpers/files/archive/TarBzip2.java
itzg: Yeah it's kind of in the realm of https://en.wikipedia.org/wiki/Composition_over_inheritance

src/main/java/me/itzg/helpers/files/archive/Archive.java
itzg: I didn't feel strongly about this one and we can just deal with it if/when it happens.

itzg:
Included in https://github.com/itzg/docker-minecraft-server/pull/4274

PR #868, opened 2026-09-23
refactor: Archive Command

harrisonablack:
After #863 was merged, i looked at the retrofit for itzg/docker-minecraft-server#4269, and realised once again we needed another feature to not regress; extracting select files. In the midst of implementing that feature i realised, for Zips and Tars, i'm effectively writing the same thing twice, using two abstractions, Java NIO and the Apache Commons.

Instead, i've simplified the two `Zip` and `TarArchive` classes into one, using only the Apache commons abstractions, and extending the `ArchiveOpener` to accept a `ArchiveInputStream`, as well as removing the interface.

I initially attempted to open this up as a [Stacked PR](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests), including the added select file abstraction, and additional documentation, but i realised all branches much exist on the original repo, not a fork :(. Instead if and when this is merged, i already have a follow up PR ready implementing the aforementioned select file abstraction.

Just a side note, I think the biggest lesson from the implementation of the zip/archive command is planning, having the forsight and looking at itzg/docker-minecraft-server$#4269 a bit longer could've definetly shrunk this down PR wise.


itzg (approved):
Good insight on the lesson learned. Code is definitely looking tighter with this one. Thanks.

itzg:
I'll wait to cut a release until after the next PR.

PR #869, opened 2026-09-23
ArchiveCommand selective extraction

harrisonablack:
Extended `ArchiveCommand` by adding List<String> parameter to the `ArchiveCommand#extract`

```text
Usage: mc-image-helper archive extract [--overwrite] ARCHIVE DESTINATION
                                       [FILE...]
Extracts an archive, checking for path traversal
      ARCHIVE       Path to archive
      DESTINATION   Output destination
      [FILE...]     Exact archive file paths to extract; extracts everything
                      when omitted
      --overwrite   Overwrite existing files
``` 

I primarily chose this way to continue passing file arguments similar to the usage in docker-minecraft-server, using "$@" 

[`unzip -o -q -d "${destDir}" "${src}" "$@"`](https://github.com/itzg/docker-minecraft-server/blob/94494ed74431d6f68203344cce8860175e9a2bca/scripts/start-utils#L622)

The selection process initially scans over the Archive, constructing a Set of all files existing. This set is then used to check if any ZipSlips occur, as well as checking that all the files the user requested actually exist within the archive. Then we iterate through the archive, extracting desired files, extending the previous extraction logic.


itzg (approved):
Nice approach. I was glad to see the validation all combined into one. And always appreciate the thorough tests.

itzg:
Oh, just remembered the README needs an update, but that can be done at a later time.

itzg:
https://github.com/itzg/mc-image-helper/releases/tag/1.70.0 building now

harrisonablack:
> Oh, just remembered the README needs an update, but that can be done at a later time. 

Yup, I've got one more PR coming, adding some javadoc and updating README

PR #870, opened 2026-09-23
Add Javadoc across Archive classes

harrisonablack:
I realised as well, due to the fact that the `extract` command is a subcommand of `Archive`, it doesn't generate usage for the subcommand.

```
Usage: mc-image-helper archive [-h] [COMMAND]
Archive helpers, checking for path traversal and extracting archives
  -h, --help
Commands:
  check-path-traversal  Checks if an archive contains path traversal attack
  extract               Extracts an archive, checking for path traversal
```

However i am happy to manually add a section after the generated usage towards the bottom of the README if that would be helpful


itzg (commented):
Thanks. Yeah, a manually created section would be good for now.

itzg (approved):
Thanks

